PR #9, opened 2025-04-23
Improve code format

Marsman1996:
Previous `xtask fmt` only formats the code in `src/*`. 
If the code file is in `src/folder/`, the `xtask fmt` will ignore it.


rikosellic:
Thank you so much!

PR #29, opened 2025-05-22
Refactored utils

rikosellic:
This pull request rebases all tree-related definitions on trace, makes some proofs more elegant and adds annotations. New codes temporarily reside in `utils2.rs`.


rikosellic:
It turns out the isomorphism between `in_subtree` and `in_subtree_range` is quite tricky and requires a bunch of properties about set and functions. I've added 'function_properties' in `vstd_extra` and prove this isomorphism.

PR #33, opened 2025-05-28
Fix rw version

Xungan2:


rikosellic:
CI/CD fails because of missing `lock-protocol/src/test/rw_version/exec/utils.rs`. Please run `cargo xtask update` to update to the latest `verus` and add the missing files. Also, remember to `cargo xtask fmt` before submitting.

PR #35, opened 2025-05-29
Refine the `rw_version` spec's invariant

rikosellic:
This pull request refines the rwlock version's state machine in `test/rw_version/spec/tree.rs`. I added some new invariants to further capture the essence of the protocol and refined some existing invariants using new definitions in `utils.rs` and `vstd_extra`. Some invariants are renamed for better readability, but the field definitions and transitions remain the same.


rikosellic:
I've added three new invariants:
- `inv_write_lock_cursors_subtree_locked`: Every write-locked node's subtree should have no readers.
- `inv_write_lock_cursors_distinct`: Different write-lock cursors should lock different nodes.
- `inv_write_lock_nodes_cursors_relation`: Each write-locked node should be write-locked by a cursor.

`inv_non_overlapping` and `inv_rc_positive` can be derived so their #[invariant] attributes are removed. I believe these modifications further capture the design principle of lock-protocol and are necessary to prove `write_lock_inductive`.

junyang-zh:
Haven't read all the changes yet. But seems good!

PR #40, opened 2025-06-03
Fix rw version

Xungan2:
Fix the rw version and reorganize files.


Xungan2:
Don't know how to handle the error "cannot call function `core::ops::deref::Deref::deref` with mode exec".

rikosellic:
You may refer to the modification in #41 or just reopen this issue, as I plan to fix it.

PR #44, opened 2025-06-05
fix the rw version and reorganize files

Xungan2:


rikosellic:
There's a pre-condition fails error in `allocate` and a Verus panic about `ManuallyDrop` in `lock_range`, marked as `#[verifier::external_body]` now.

PR #76, opened 2025-07-12
fix some lemmas in ghost_tree, update fvt11

WeituoDAI:
Thanks for sharing this interesting project.

In this PR, I update the following things:

- vstd_extra/ghost_tree.rs
  + there are several incorrect lemmas (about trace), I fix them
  + the def of `trace` seems incorrect, I modify it
  + add other lemmas

- fvt11
  + update the proof due to the change of vstd_extra

### DIFF
--- a/fvt11-pt-cursor-guards/src/page_table/cursor/mod.rs
+++ b/fvt11-pt-cursor-guards/src/page_table/cursor/mod.rs
@@ -68,7 +68,7 @@ impl<'a, M: PageTableMode> Cursor<'a, M> {
         barrier_lv == model.locked_subtree@.level && {
             forall|i: int|
                 0 <= i < barrier_lv ==> locked_path.index(i) == barrier_start_path@.index(i)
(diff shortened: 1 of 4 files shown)


rikosellic:
Thank you very much for the contribution! Indeed, the `ghost_tree` library is incomplete and not carefully inspected. That's why we avoid using it in `lock-protocol`, our current active development for our SOSP submission. Development on `fvt`s is currently on pause because the codebase to be verified has evolved significantly, and we are planning a major verification framework overhaul.

If you are interested in contributing further, a good target may be the `NodeHelper` in `lock-protocol/src/spec/utils.rs`, which serves as another tree library in `lock-protocol`. It shares some overlap with `ghost-tree`, and should be extracted or merged with `ghost-tree` as an independent library, but we have not thoroughly thought about how to do this.

Thanks again for your time and valuable input—please feel free to reach out if you’d like more context or collaboration.

PR #83, opened 2025-07-15
Verify `PagingConstsTrait`

rikosellic:
- Add `lemma_consts_properties` in `PagingConstsTrait`
- Prove `lemma_consts_properties` for `PaginGConsts`
- Use the correct definition of `page_size`, `nr_subpage_per_huge`.

@zouyonghao Please let me know if there are any conflicts.


rikosellic:
I realize that it is not finished yet, converting it into a draft.

PR #87, opened 2025-07-16
Refactor `MetaSlot` in `rw`

rikosellic:
This PR refactors `MetaSlot` to use Asterinas-style (partially implemented so far).  It may serve as a walkaround for `dyn` type. The `usage` information is moved into the ghost `MetaSlotPerm` as a verification-only parameter. Currently, only the independent definition in `rw` is modified.


rikosellic:
@zouyonghao @Xungan2 any suggestions?

PR #117, opened 2025-07-25
Support replace_with_none

zouyonghao:
Signed-off-by: Yonghao Zou <zouyonghao@live.cn>


lock-protocol/src/mm/page_table/cursor/spec_helpers.rs
```diff
@@ -81,4 +81,37 @@ pub open spec fn spt_do_not_remove_above_level<C: PageTableConfig>(
         }
 }
 
+pub open spec fn spt_do_not_change<C: PageTableConfig>(
```
junyang-zh, line 84: Seems just do `spt == old(spt)`.

PR #142, opened 2025-07-29
Make pte_index refer to C for constants instead of PagingConsts

Deophius:
In the `page_size` function queries its generic parameter `C` for constants instead of `PagingConsts`, but the `pte_index` uses `PagingConsts`. I think this inconsistency is a bug, so I changed `pte_index` to use `C` instead of `PagingConsts`.

Also contains some minor fixes to the proof of `move_forward`.


zouyonghao:
LGTM!

PR #151, opened 2025-07-30
Add proof for page_size_spec properties

Deophius:
There was proof for properties of the exec fn `page_size`, but I think adding properties to the corresponding spec fn is helpful in verifying `move_forward`.


rikosellic:
LGTM. Thanks!

PR #171, opened 2025-08-01
WIP：Fix rcu tree spec

rikosellic:


lock-protocol/src/spec/rcu/tree.rs
```diff
@@ -147,18 +152,30 @@ pub fn inv_stray_has_false_implies_pte_is_alive(&self) -> bool {
...
-                                    .filter(|pair: ((NodeId, Paddr), bool)| pair.1 == false)
-                                    .len() == 0
+        NodeHelper::valid_nid(nid) && self.nodes[nid] is Free ==>
+        {
+            self.strays_filter(nid).len() == 0
         }
+}*/
+
+#[invariant]
+pub fn inv_stray_in_node(&self) -> bool {
```
Xungan2, line 174: The invariant is wrong. 
Every node has a stray flag, which is stored in the corresponding metaslot. The stray flag indicates whether the node exists in the page table. True implies the node is removed and waiting for recycling. False implies the node is in the page table.
The rcu tree spec model node states of a page table. Thus, every page table node should have a false stray flag. Also,  the paddr stored in the stray enum should be consistent with the address of the node. 
I think `inv_pte_is_alive_implies_stray_has_false` and `inv_stray_has_false_implies_pte_is_alive` are enough to describe these properties.

rikosellic: Thanks for the clarification. I recognize this invariant is indeed wrong. But in `protocol_allocate` and `normal_allocate`, the `add strays += [ (nid, paddr) => false ];` step cannot be established because we only know `nid` is not in the page table tree, but it may be marked stray at the same paddr and is waiting to be recycled. Therefore, allocating to the same address is illegal. Maybe what we need is to add another requirement in these transitions that the paddr is distinct from all existing paddr associated with `nid`. Do you think such a modification is reasonable?

rikosellic:
- Add a `free_paddrs` field.
- Add `inv_free_paddr_not_in_strays` that states every paddr in `strays` is not free.
- Delete a lot of proofs because Verus can solve them automaitcally.

Xungan2:
You remind me that the cursor mechanisms of rw and rcu are different. The `inv_non_overlapping` essentially guarantees that all write-locked nodes should be exclusively owned. Compared with the rw version, if the cursor of rcu is in state `Locking(nid1, nid2)`, then all nodes between `[nid1, nid2)` are write-locked. So we should make some changes to `inv_non_overlapping`.

PR #175, opened 2025-08-04
Prove rcu tree spec

rikosellic:


rikosellic:
@Xungan2  It seems there is a fundamental error in `protocol_lock_start`. At the end of the traverse phase, we do lock the subtree root if it is not stray, so `protocol_lock_start` will create a locked cursor `Locking(nid,nid+1)` instead of locking nothing `Locking(nid, nid)`. I think a correct `protocol_lock_start` should be something like 
```
transition! {
    protocol_lock_start(cpu: CpuId, nid: NodeId) {
        require(valid_cpu(pre.cpu_num, cpu));
        require(NodeHelper::valid_nid(nid));

        remove nodes -= [ nid => NodeState::LockedOutside ];
        remove cursors -= [ cpu => CursorState::Void ];
        add nodes += [nid => NodeState::Locked];
        add cursors += [ cpu => CursorState::Locking(nid, nid+1) ];
    }
}
```
Note that it changes a `LockedOutside` node in the traverse phase into a `Locked` node in the locking phase. Perhaps we should add another requirement regarding the `strays` condition in the transition, and also modify the well-formedness of `CursorState::Locking` to ensure that the root is both present and locked.

rikosellic:
- Add `inv_cursor_root_in_nodes`: Every non-empty cursor's root should be locked.
- Add `inv_not_allocated_subtree`: Every unallocated node's descendants can not be allocated.
- Add `inv_locked_node_state`: Every node in the range of cursor should either be locked or not allocated.
- Add `inv_inv_unallocated_node_locked_implies_in_subtree`: The other direction of `inv_locked_node_state`.
- Use `Set` to represent `locked_range`
- Add the `cpu` parameter in `protocol_allocate`, `protocol_deallocate`.
- Other minor modifications in transitions.

PR #183, opened 2025-08-08
Verify Cursor::move_forward

Deophius:
Verified `Cursor::move_forward`.

Added two alternative `spec fn` for `pte_index` that doesn't use bit arithmetics and some properties for the `page_size` function.

Fixed a typo in `lemma_aligned_pte_index_unchanged`.


zouyonghao:
Thank you! This is very impressive.

PR #202, opened 2025-09-26
Prove `assume` in Entry::replace

Marsman1996:


Marsman1996:
One bad news is that LLM would add exec code sometimes, and in this case it adds
```rust
let pte_addr: usize = self.node.paddr() + self.idx * exec::SIZEOF_PAGETABLEENTRY;
self.pte.set_paddr(pte_addr);
```
I'm working on this...

PR #203, opened 2025-09-28
PhaseII verified: Merge `vstd-extra` from `main`

rikosellic:
Merge `vstd-extra` from main, bringing the following changes to align with the latest Verus:
- Use `uninterp`, `axiom` keywords
- Fix `when_used_as_spec` use in traits
- Add `function_properties`, `set_extra`, `map_extra`
- Prove and add lemmas in `ghost_tree`, `seq_extra`


SNoAnd:
Looks good to me.

PR #214, opened 2025-10-23
PhaseII verified: Remove reference in `inv` signature

rikosellic:

PR #218, opened 2025-10-27
PhaseII verified: Use `self` in `Inv`

rikosellic:
This is a redo of #214. I removed all `&`s in `Inv`, `InView`, and `OwnerOf` while adding `Sized` as a constraint.


SNoAnd (approved):
This is cleaner, and doesn't change anything logical, so I'm happy with it.

PR #246, opened 2025-12-10
prove: admit in unlock_range

Marsman1996:
prove the `admit` inside `unlock_range`, and  simplify proof code by removing `assert`


rikosellic:
LGTM！The simplification function makes me very happy.

PR #247, opened 2025-12-11
refactor: simplify unnecessary assert

Marsman1996:
Updated: 
Now I filter the unproven functions (i.e., functions with `external_body `, `admit` and `assume`)

|Status|Spec | Proof | Exec|
|----- |---- | ----- | --- |
|Before|5192 |  4787 | 1754|
|After |5192 |  4471 | 1754|

Simplify 316 (6.6%) lines of proof code

---

|Status|Spec | Proof | Exec|
|----- |---- | ----- | --- |
|Before|5192 |  4787 | 1754|
|After |5192 |  4229 | 1754|

Simplify 558 (11.7%) lines of proof code

### DIFF
--- a/.github/workflows/format-verify.yml
+++ b/.github/workflows/format-verify.yml
@@ -48,7 +48,7 @@ jobs:
             echo "Code is not properly formatted. Run 'make fmt'."
             git diff
             echo "FMT_FAILED=1" >> "$GITHUB_ENV"
-            # exit 1
+            exit 1
           fi
 
       - name: Publish summary


Marsman1996:
> One small question: does the simplification only touch code that has already been proved (i.e., with no `admit`, `assume`, or `external_body`)?

Well, I forget to skip the function marked with `external_body`. 
However, how to avoid the `admit`, `assume`? Should we skip the funtion containing one of them?

PR #248, opened 2025-12-11
prove: `admit` in `to_ref`

Marsman1996:
I’m using a proof agent to prove some `admit` statements in the code. However, for `to_ref`, it modifies the specifications.


rikosellic:
I'm not very confident about this change. It adds a new permission for input, but does not tell us where we can produce such a permission. Not familiar enough with this part, though, I will look into it in more details later.

SNoAnd:
I have `Cursor::query` verified in my local branch, and this change should be easy to integrate. In the verified version the parent `NodeOwner` is already proven to be present and well-formed in order for the `guard_perm` to be available.

I'll aim to have the merge up tomorrow.

PR #255, opened 2025-12-18
prove: `admit` in linked_list

Marsman1996:
This is generated by the proof agent, but the agnet modify the exec code(?)

### DIFF
--- a/ostd/src/mm/frame/linked_list.rs
+++ b/ostd/src/mm/frame/linked_list.rs
@@ -272,16 +272,10 @@ impl<M: AnyFrameMeta + Repr<MetaSlot>> LinkedList<M> {
             return None;
         };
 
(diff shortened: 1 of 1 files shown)


rikosellic:
I think this modification is somehow closer to the Asterinas source code. It corresponds to the get_slot method in Asterinas, which obtains a reference of the  MetaSlot from a Paddr. The original verification code, however, uses the take method to exclusively own the MetaSlot and puts it back. Although these two methods are equivalent in an isolated sequential context, it is actually impossible in the real system becuase of the concurrent access. What do you think @SNoAnd ?

SNoAnd:
Looks great! In general, I prefer borrow over 'take-put', so I think this change is perfectly valid. I don't remember why this was in 'take-put' style in the first place, but it doesn't seem necessary anymore.

PR #281, opened 2026-01-13
More Resource Algebra

rikosellic:


rikosellic:
By now, there are the following resource algebras:
## PCM
- `ExclR`: exclusive ownership, standard in Iris.
- `AgreeR`: duplicable agreement ownership, standard in Iris.
- `CsumR`: the sum type, standard in Iris.
- `CountR`: an `Auth` counter produces and recycles tokens and maintains the number of existing tokens, mimicking a reference counter.

## Storage Protocol Monoid
- `FracS` and `FracStorage`: a real-number-based fractional ownership, which is a re-implementation of `FracCarrierOpt` and `FracGhost` in `vstd::tokens::frac` that supports arbitrary finite separation instead of a compile-time constant.
- `hybrid_product`: a `HybridProduct` of a PCM and a protocol monoid produces a protocol monoid.

Currently, I'm investigating whether `HybridProduct<CountR, FracS<T>>` can serve as the model of `RwLock`.

@zouyonghao @Xungan2  Referencing you in case you are interested.

zouyonghao:
This is very interesting. What's the benefit of the real-number-based fractional ownership?

rikosellic:
In Iris, fractional ownership is implemented using positive rational numbers, allowing shared read access to be split among multiple users. When the fractions sum to one, the current holder has full ownership and is permitted to update the resource. `FracGhost`, however, implements it with `TOTAL:usize`, which requires prior knowledge of the maximum number of concurrent read accesses at compile time. As Verus recently supported real numbers in spec, now we can implement it with real-number fractions to split an arbitrary number of times at runtime. The exact numerical value of each fraction is irrelevant.

It may seem that a simple `&` will also provide similar shared access. However, fractional ownership has several important advantages:
- At the theoretical level, the meaning of a `&` to a ghost resource is not straightforward; fractional ownership provides a clear and explicit model of shared permissions.
- `&T` only dereferences to `T`, the Storage Protocol Monoid has a `guard` operator to obtain a reference of a different type.
- Fractional ownership composes naturally with other PCMs, enabling more sophisticated resource models.

zouyonghao:
I agree with your opinions, but I think Verus's ownership model is enough to prove most of the targets, like RwLock. (or no?)

PR #296, opened 2026-02-03
prove: 6 admits in ghost_tree

Marsman1996:


rikosellic:
@Marsman1996 I've simplified and proved the insert case, is it possible for KVerus to simplify the remove case?

vstd_extra/src/ghost_tree.rs
```diff
@@ -1483,31 +1515,105 @@ impl<T: TreeNodeValue<L>, const N: usize, const L: usize> Node<T, N, L> {
         }
     }
 
+    /// Check whether remove satisfies rel_children at the removal point
+    pub open spec fn remove_rel_children(self, path: TreePath<N>) -> bool
```
rikosellic: This function is unnecessary, it can be expressed using recursive_trace, like in lemma_recursive_insert_preserves_inv.

PR #303, opened 2026-02-09
RwLock

rikosellic:
KVerus Prompt:
You need to add the type invariant of `RWLock` in `ostd\src\sync\rwlock.rs`, which is defined in the function `closed spec fn wf(self) -> bool`. 

The `RwLock` structure's `lock` field has the type of `lock: AtomicUsize<_, Option<RwFrac<T>>,_>,`, which extends the ordinary atomic `usize` with a ghost `RwFrac` permission. The `RWFrac` type is an alias of type `Frac<pcell::PointsTo<T>,MAX_READER_U64>`. `Frac` is the classical fraction permission, whose definition can be found in `vstd/tokens/frac.rs`; `pcell::PointsTo` is the permission to access a cell, which is defined in `vstd/cell/pcell.rs`.

You should analyze the behaviour of the `SpinLock` methods (commented in the same file), including `new, read, write, upread, try_read, try_write, try_upread`, and write the correct invariant between the `AtomicaUsize`'s value and the ghost `RwFrac` permission. To learn the syntax of the invariant and how it is used, you can refer to the `SpinLock` in`ostd\src\sync\spin.rs` as an example.

### DIFF
--- a/ostd/src/sync/rwlock.rs
+++ b/ostd/src/sync/rwlock.rs
@@ -829,23 +829,6 @@ impl<'a, T /*: ?Sized*/, G: SpinGuardian> RwLockUpgradeableGuard<'a, T, G>
             ghost g => {
                 lemma_consts_properties_prev_next(prev, next);
                 if res is Ok {
-                    assert(prev & WRITER == 0usize) by (bit_vector)
-                        requires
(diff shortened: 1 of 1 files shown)


rikosellic:
Very interesting, I will check it later.

PR #312, opened 2026-02-14
Documentation progress

SNoAnd:
This PR contains documentation for `LinkedList` and and some for `VmSpace`, though I still have more to do on the latter. There is also new documentation in `MetaRegionOwners` that relates to the meta slot model and its memory safety, which is what I'll be focused on next, because that is most closely connected to the question of what guarantees we can make about ill-behaved (but safe Rust) callers.


ostd/src/mm/frame/linked_list.rs
```diff
@@ -53,6 +55,19 @@ pub struct Link<M: AnyFrameMeta + Repr<MetaSlot>> {
 ///     [`from_in_use`] a frame that is inside a linked list.
 ///  3. We also allow creating cursors at a specific frame, allowing $O(1)$
 ///     removal without iterating through the list at a cost of some checks.
+/// # Verified Properties
+/// ## Verification Design
+/// The linked list is abstractly represented by a
+/// [`LinkedListOwner`](crate::specs::mm::frame::linked_list::linked_list_owners::LinkedListOwner)
+/// which is a sequence of [`LinkOwner`](crate::specs::mm::frame::linked_list::linked_list_owners::LinkOwner)
+/// objects. Each [`LinkOwner`] uniquely owns the frame metadata of a link.
+/// ## Invariant
```
rikosellic: Can we post the Verus definition of the invariant (or key parts of it) here so that readers can directly compare the formal definition with the natural language descriptions, like what we've done in SpinLock.

ostd/src/mm/vm_space.rs
```diff
@@ -1051,9 +1054,21 @@ impl<'rcu, A: InAtomicMode> Cursor<'rcu, A> {
     /// Otherwise, it will return `None`. And the cursor may stop at any
     /// address after `len` bytes.
     ///
-    /// # Panics
-    ///
-    /// Panics if the length is longer than the remaining range of the cursor.
+    /// # Verified Properties
+    /// ## Preconditions
+    /// The cursor must be within the locked range and below the guard level.
```
rikosellic: For multiple pre/post conditions like this, perhaps we can use bullets to make the doc clearer.

rikosellic:
Thanks so much! It generally looks good to me, left a few comments on the documentation.

rikosellic (changes requested):

ostd/specs/mm/frame/meta_region_owners.rs
```diff
@@ -24,6 +24,24 @@ verus! {
 /// Cell<MetaSlot> at a fixed address range.
 pub struct MetaRegion;
 
+/// Represents the ownership of the meta-frame memory region.
+/// # Verification Strategy
```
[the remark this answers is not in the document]
SNoAnd: Done!

rikosellic (approved):
LGTM now! Thank you for your great effort!

PR #315, opened 2026-02-23
`MetaSlot` documentation and permissions refactor and

SNoAnd:
This PR primarily covers the inline docs for `MetaSlot`. For increased readability I also refactored the structure of permissions for the slots and their fields. Tracking of raw frames has been made much more visible, for instance.

I've been doing my best to prune extraneous preconditions and clearly mark pre- and post- conditions with whether they impact safety or correctness, though that is less essential here because `MetaSlot` is not public. I prioritized documenting them because, while they aren't part of the public API, they are pretty crucial for understanding the nature of the soundness claims, because they hold much of the unsafe code.


rikosellic:
Thanks for the hard work！LGTM and I will investigate the verification design in the next few days.

By the way, `cell::PCell` is now deprecated, can you replace it with the new `cell::pcell::PCell`?

SNoAnd:
I will work on the PCell replacement, yeah!

PR #329, opened 2026-02-27
`VmSpace` API fix

hiroki-chen:
As discussed on Zulip, we reverted the changes to exec-mode `VmSpace` and leave all other verification APIs to `VmSpaceOnwer` in `ostd::specs:mm:vm_space::VmSpaceOwner`. `create_writer` and `create_reader` are renamed back to `writer` and `reader` respectively.


rikosellic:
LGTM! Thanks!

PR #335, opened 2026-03-05
Add documentation for `vm_space` utilities

hiroki-chen:
This should cover some low-level verified primitives for playing with `vm_space` stuff.


rikosellic:
LGTM, thanks!

PR #346, opened 2026-03-12
Infrastructure for `sync::rcu::non_null::either`

Marsman1996:
Cannot prove `raw_as_ref` and `from_raw` without modifying a lot to the other part of the code...


rikosellic:
The current nonnull proof design is fairly restrictive, becuase of the smartpointsto deisgn. I'm thinking maybe we can use a trait instead of a enum. But I have not decided yet.

rikosellic:
@Marsman1996 I've implemented the trait design and refactor the verification in #414. Can you update this PR after it is merged? Thanks!

Marsman1996:
What's the diff when the PR is merged

rikosellic (approved):
Although `impl NonNullPtr for Either` is still not proved, I think we have accumulated many useful results that deserve to be merged.

PR #348, opened 2026-03-17
A new sum-type logical resource library

rikosellic:
This PR adds a new sum-type logical resource library based on the Leaf storage protocol, originally planned in #340, to replace the extra `read_guard_token` field in `RwPerms`.
### Concept
The logical resource `SumResource<A,B,TOTAL>` can store either a `tracked A` or a `tracked B` object.  Then, assuming a `tracked A` is stored, the resource can be split up to `TOTAL` pieces. The fancy part is that only one piece has the exclusive ownership of the `tracked A`, so it can take the storage out of the resource by a `take_resource` method, much like `Frac::take_resource`, push it back, update it, or borrow it using `tracked_borrow`. The other pieces can not obtain the resource, but they share the knowledge that a resource of type `A` may exist, so one can reason that any resource of type `B` can not exist.

The `change_to_left/right` part is a bit more restrictive. One cannot simply obtain a single piece of type `A` and replace it with type `B`, because resources elsewhere may still infer that a resource of type A exists. But if one can gain all the `TOTAL` pieces to confirm that there are no other pieces left, then we can update the resource with whatever object we want.

### Implementation
This feature is implemented with the following protocol monoid:
```rust
/// The knowledge of the existence of a specific type of resource can be shared up to TOTAL pieces,
/// but only one piece has the exclusive ownership of the resource,
/// allowing arbitrary withdrawing, depositing, or updating.
pub ghost enum CsumP<A, B, const TOTAL: u64> {
    /// The unit element, only for technical reasons, not intended to be used directly.
    Unit,
    /// The left side of the sum, with an optional resource, a fraction, and a boolean indicating whether it is the exclusive owner of the resource.
    Cinl(Option<A>, int, bool),
    /// The right side of the sum, with an optional resource, a fraction, and a boolean indicating whether it is the exclusive owner of the resource.
    Cinr(Option<B>, int, bool),
    /// An invalid state, used to represent an invalid combination of resources.
    CsumInvalid,
}

impl<A, B, const TOTAL: u64> Protocol<(), Sum<A, B>> for CsumP<A, B, TOTAL> {
    open spec fn op(self, other: Self) -> Self {
        match (self, other) {
            (CsumP::Unit, x) => x,
            (x, CsumP::Unit) => x,
            (CsumP::Cinl(ov1, n1, b1), CsumP::Cinl(ov2, n2, b2)) => {
                if !self.is_valid() || !other.is_valid() || n1 + n2 > TOTAL || b1 && b2
                    || ov1 is Some && ov2 is Some {
                    CsumP::CsumInvalid
                } else {
                    CsumP::Cinl(
                        if ov1 is Some {
                            ov1
                        } else {
                            ov2
                        },
                        n1 + n2,
                        b1 || b2,
                    )
                }
            },
            (CsumP::Cinr(ov1, n1, b1), CsumP::Cinr(ov2, n2, b2)) => {
                if !self.is_valid() || !other.is_valid() || n1 + n2 > TOTAL || b1 && b2
                    || ov1 is Some && ov2 is Some {
                    CsumP::CsumInvalid
                } else {
                    CsumP::Cinr(
                        if ov1 is Some {
                            ov1
                        } else {
                            ov2
                        },
                        n1 + n2,
                        b1 || b2,
                    )
                }
            },
            _ => CsumP::CsumInvalid,
        }
    }
```

### DIFF
--- a/ostd/src/sync/rwlock.rs
+++ b/ostd/src/sync/rwlock.rs
@@ -67,6 +67,19 @@ exec const V_MAX_READ_RETRACT_FRACS: u64
     (MAX_READER_MASK + 1) as u64
 }
 
+spec const V_MAX_READ_GUARDS_SPEC: u64 = MAX_READER as u64;
+
+#[verifier::when_used_as_spec(V_MAX_READ_GUARDS_SPEC)]
+exec const V_MAX_READ_GUARDS: u64
+    ensures
+        V_MAX_READ_GUARDS == V_MAX_READ_GUARDS_SPEC,
+        V_MAX_READ_GUARDS == MAX_READER,
+        V_MAX_READ_GUARDS < u64::MAX,
+{
+    assert(MAX_READER < u64::MAX) by (compute_only);
+    MAX_READER as u64
+}
+
 tracked struct RwPerms<T> {
     /// The permission of the `PCell<T>`. 
     /// In the reading state, it can be splited up to `V_MAX_PERM_FRACS:= MAX_READER + 2` pieces,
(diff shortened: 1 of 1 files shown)


zouyonghao:
This is interesting, and hopefully, there is an example for using the sum-type.

PR #356, opened 2026-03-23
prove: sync::mutex

Marsman1996:


ostd/src/sync/mutex.rs
```diff
@@ -1,165 +1,300 @@
...
     ops::{Deref, DerefMut},
-    sync::atomic::{AtomicBool, Ordering},
+    sync::atomic::Ordering,
 };
 
 use super::WaitQueue;
 
+verus! {
+
+tracked struct MutexPerms<T> {
```
rikosellic: There is no need to create another struct since there is only one field.

rikosellic (changes requested):

rikosellic (approved):

PR #372, opened 2026-03-26
Fix missing `deref` specifications

hiroki-chen:
This PR complements #358 as doing `rwlock_guard.inner_field` calls `deref` implicitly which would require us to have a `view` available there. So I added `returns self.view()` in case that anyone uses it in the future.


hiroki-chen:
BTW, other lock guards might also need to define this. Should be a trivial change I guess.

Issue #344, opened 2026-03-10
VMIO trait implementation for other components


hiroki-chen:
I'm starting to provide the verified APIs (to some extent, specs) for components that utilize VMIOs for reading/writing. This should be the issue for tracking the progress.

hiroki-chen:
The lifetime marker is meaningless at runtime so verification would block the use of `RwArc` and will contaminate the struct. Perhaps we need to use the resource-based reasoning.

hiroki-chen:
#390 Partially fulfills `dma`, let's see what can be further implemented.

PR #390, opened 2026-03-30
Experiment: verification for `dma` module

hiroki-chen:
This partially fulfills #344's goal but since `RwArc` is a runtime-checked object, we define most of the APIs on the `RwLock*Guard` which as a view method on it. 

The most significant changes include the introduction of a new DMA mapping set infrastructure, enhanced specification and verification for untyped memory segment access, and a substantial redesign of the `VmIo` trait to unify and clarify read/write specifications. Additionally, the PR includes various code cleanups, import reorderings, and minor improvements to frame and segment APIs.

### DMA Mapping Infrastructure

Basically introduced verified APIs for DMA operations.

### Untyped Memory Segment Access

- Implemented verified `reader` and `writer` methods for `Segment<M>` where `M: AnyUFrameMeta + OwnerOf`, with detailed specification and proof blocks to ensure correct kernel-space access and invariants.

### Virtual Memory I/O Trait Redesign

- Refactored the `VmIo` trait to unify and clarify read/write specifications, replacing multiple `*_requires` and `*_ensures` spec functions with `read_spec` and `write_spec`, and updating the method signatures for `read` and `write` to use `VmWriter`/`VmReader` and ownership tokens. This makes the contract for virtual memory I/O more explicit and verifiable.

I'm not sure if the usage of `RwArc` and some of the derived structs make sense so any comment or advice is strongly welcomed.


hiroki-chen:
@rikosellic This is ready for merge :)

hiroki-chen:
Oh BTW the inner `USegment` has been removed and replaced with explicit type parameters due to the poorly supported trait objects.

PR #392, opened 2026-04-01
Fix warnings

Marsman1996:
~This time I add a spec modfication evaluation module and the agent drops the specs whose rating is lower than 7~

A warning fix PR now. Thanks for the effort of @SNoAnd for estimating the `admit()`s in `mm`.


rikosellic:
Reviewing this requires a deep understanding of the current MetaSlot model, which, unfortunately, I do not have yet. Need some time.

rikosellic:
@SNoAnd @Marsman1996 Is this PR still valid?

Marsman1996:
> @SNoAnd @Marsman1996 Is this PR still valid?

Nope. 
I will update the PR, maybe in 5.2 or later.

Marsman1996:
@rikosellic 
Shall we just comment the `#![warn(missing_docs)]` in `ostd/src/lib`? 
Since it might be unnessary to add doc for `type_inv` or something like that.

Marsman1996:
The lifetime warning fix is align with the latest Asterinas code
https://github.com/asterinas/asterinas/blob/5089f17d1facbd80ebedfdef5e119e91e2100b0d/ostd/src/sync/rwarc.rs#L44

rikosellic:
Yeah I think we can suppress the doc warnings.

Marsman1996:
There are several `unsafe` are missing, such as the `from_raw` in `ostd/src/mm/frame/mod.rs`.
Don't know why...

rikosellic (approved):

rikosellic:
Thanks for your contribution！

PR #403, opened 2026-04-13
Remove `addr` field in `ReprPtr`

rikosellic:
The `addr: usize` field in `ReprPtr` does not provide any new information, as the `ptr: PPtr` field already has an `addr` method that produces a `usize`. This PR deletes this field, saves a bit line, and it may be easier to migrate if Verus supports `dyn` better in the future. @SNoAnd, do you think this change is okay?


rikosellic:
The `take_inner_perms/put_inner_perms/take_pointsto` methods also seem unsound and should be replaced with the `new-mut-ref` support. However, I ran into some Verus bugs when trying these modifications. I've created an issue in Verus Zulip and I hope there will be response from Verus maintainers soon.

rikosellic:
> The `take_inner_perms/put_inner_perms/take_pointsto` methods also seem unsound and should be replaced with the `new-mut-ref` support. 

@SNoAnd I find that we no longer need any helper functions like these anymore, as we can directly use `&mut` in `ostd` APIs, and it works pretty well!

SNoAnd:
Yes, this looks great! The `take_inner/put_inner` axioms are definitely worth replacing with borrows now that they're supported. Adding `FromSpecImpl` is also a good idea.

PR #408, opened 2026-04-14
Make `PageTableConfig::TOP_LEVEL_CAN_UNMAP` be associated const

rikosellic:
Verus's support for associated const is limited, but is powerful enough to support `PageTableConfig::TOP_LEVEL_CAN_UNMAP`. This PR serves as the first example to use associated const.


rikosellic:
It seems this change siginificantly slows down the verification. Maybe we should wait until Verus's support for associated const is more mature.

rikosellic:
I experimented yesterday, and the associated constant is still not good enough.

PR #410, opened 2026-04-15
Better proof support for duplicate frames, which closes more admits

SNoAnd:
This PR overhauls to the frame tracking in the page table to deal with cases where a single frame is mapped multiple times. Previously we tracked a single optional path, `path_if_in_pt` for each frame; now it's a set of paths, `paths_in_pt`. This enables that tracking to be used more robustly in a number of proofs, so it also closes more admits, and should make the ones related to `split_if_mapped_huge` much easier.


rikosellic:
Well done, thanks!

PR #419, opened 2026-04-20
Fill API gaps and admits

SNoAnd:
This PR is a bunch of smaller stuff. There were some missing trait functions that are now added and verified, and a number of admits in the page table lemmas that are closed. The deepest structural change is that the connection between a node's tracked path and its index within its parent's children is now tracked by a separate invariant that is not included in `Node::rel_children()`, because it wasn't possible to prove `default_preserves_rel_children` with that condition. But that's just changing how we track the invariant, not changing the invariant itself.


rikosellic:
Thanks for the contribution!

PR #420, opened 2026-04-22
Panic/assert rework

SNoAnd:
This PR deals with how we treat panics and assertions. Previously we had some unsatisfying workarounds because Verus has a hard time reasoning about panics. Now, the function `panic_diverge` from `vstd_entra::panic.rs` will tell Verus that the current code path diverges. If called in a conditional branch or loop body, Verus can deduce that the conditional is false. I've added `assert!` and `assert_eq!` macros that use this ability to mimic regular Rust asserts.

That means that the old `!panic_condition` precondition pattern is obsolete. Instead, since we often rely on the knowledge that we have passed an assert without panicking, I've been mechanically shifting panic conditions from the preconditions of their functions to their post conditions ("because the function terminated, these conditions must have held".) There are a few cases left where the logic is more complicated and I'm still working on making the switch, so stay tuned for more.

This change illuminates circumstances where integers may overflow prior to being used in the assertions. For now we assume that they don't overflow in order to not break the verification, and mark them as bugs.

A side feature of this PR: some closed admits in the virtual address arithmetic, handling the higher order bits of kernel vaddrs. I wanted to make sure it wasn't possible to rule out some of these overflows by tightening cursor invariants.


rikosellic:
This is very interesting! I think the new `{P}c{Q & !panic}` design is conceptually equivalent to adding a "panic-exit" clause to the Hoare-triple, like `{P}c{Q}{panic}`, is this correct?

**Update**: I realize the panic clause describes the program state *just before* the crash instead of the state *right after* the panic. It seems like a way to encode the "panic" Hoare logic into the normal Hoare logic. I think this is like a basic version of the ideas explored in the SOSP15 Crash Hoare Logic paper and OSDI25 Power paper. Anyway, just some random program logic thoughts, this PR LGTM!

SNoAnd:
@rikosellic That's an interesting paper! It goes much further than this in terms of reasoning about post-crash recovery. That would be equivalent to us reasoning about stack unwinding and safe dropping of objects, which Verus definitely doesn't give us machinery for. Instead it's a sort of reverse-temporal logic: by reaching a point in the execution, we can deduce facts about the execution's history. That means that the caller context gets the condition in exactly the same form that it did when it was a precondition. It also serves as a verified witness for the panic clauses in the doc comments, in contrapositive form.

Formally we're modeling panics as divergence and treating asserts similarly to the while rule in partial Hoare Logic: `{P} while E do C { P & !E }`. In fact one of my tests was making the `assert!` macro expand to `while(...) { assume(false) }`. Verus normally corresponds to total Hoare logic, although the `#![verifier::exec_allows_no_decreases_clause]` drops it back to partial HL.

PR #424, opened 2026-04-27
Interrelated arithmetic, &mut porting, and panic handling fixes

SNoAnd:
Lots going on in this PR. `PageTableGuard` uses &mut again, which simplifies our ability to check code equivalence. I turned some arithmetic axioms into verified proofs, which had knock-on effects at their call sites as some of them had to be split into different cases.

I used the panic handling changes to improve the `kvirt_area` specs and eliminate unnecessary preconditions, and did the `panic_cond` propagation in the remaining places that I hadn't in the last PR. That's for all of the explicit `panic_cond` predicates, though, I still need to go through `segment.rs` and `io.rs` because some of those functions followed the pattern without obeying the naming convention. The `kvirt_area` fixes also involved more precise modeling of untracked frames.


rikosellic:
@SNoAnd We have also encountered some build errors earlier today. Maybe you can try `#[verifier::spinoff_prover]`.

Marsman1996:
Why every time there are so many commits in PR 👀

SNoAnd:
@Marsman1996: Yeah, it's annoying, the PR seems to take every commit since we migrated to main. I'm not sure why, maybe a github configuration issue. The actual delta here starts with 3ac88a705.

Marsman1996:
> @Marsman1996: Yeah, it's annoying, the PR seems to take every commit since we migrated to main. I'm not sure why, maybe a github configuration issue. The actual delta here starts with [3ac88a7](https://github.com/asterinas/vostd/commit/3ac88a705be5de56a2ed087a514ea0cba57fcb17).

I think this is likely caused by the commit history difference, since changes on the `main` branch are merged via squash-and-merge.

To avoid this issue, we should probably keep `phaseII/verified` synced with the latest main branch. 
Alternatively, next time we can fork from the VOSTD `main` branch directly and open the PR from that branch.

ostd/specs/mm/page_table/cursor/page_size_lemmas.rs
rikosellic, line 1: #[verifier::spinoff_prover] will eliminate these codes.

rikosellic:
I see a major code change in `dma`. Is this consistent with the previous PR by @hiroki-chen? @SNoAnd

hiroki-chen:
Unfortunately not, the `dma_stream` in this PR seems outdated. @SNoAnd, could you please update it?

rikosellic:
@SNoAnd @hiroki-chen That was a close call, the issue almost caused serious consequences. I suggest that we always create a new branch from main each time, instead of continuing to reuse the old verified branch.

ostd/src/mm/dma/dma_stream.rs
rikosellic, line 1: This file is outdated.

rikosellic (changes requested):

hiroki-chen:
> @SNoAnd @hiroki-chen That was a close call, the issue almost caused serious consequences. I suggest that we always create a new branch from main each time, instead of continuing to reuse the old verified branch.

I think we may delete this branch after this one has been fixed?

ostd/src/mm/dma/dma_stream.rs
hiroki-chen: The dependencies should also be updated

Marsman1996:
> > @SNoAnd @hiroki-chen That was a close call, the issue almost caused serious consequences. I suggest that we always create a new branch from main each time, instead of continuing to reuse the old verified branch.
> 
> I think we may delete this branch after this one has been fixed?

Yes, I think so. We shall fork from the VOSTD main branch directly in our own repo and open the PR from that branch.

rikosellic:
Actually, I've deleted this branch before, right after https://github.com/asterinas/vostd/pull/295. It seems the branch was later pushed back to the remote repository from a local machine.

SNoAnd:
OK, I'll fix the dma issue and then also switch to using the main branch in our repo as the source for PRs.

SNoAnd:
DMA is restored from 3716ba9d. It builds fine; @hiroki-chen, please give it a quick look to make sure nothing is missing. This push also contains more admit/assume closures, and work on the segment and io transcription issues, though the latter are not fully done.

SNoAnd:
And sorry about the branch mixup!

hiroki-chen:
I took a look and this seems fine to me

rikosellic:
@SNoAnd, Can you put assumptions in `std_extra` into modules of `vstd_extra::external`? This can make things clearer, thanks!

SNoAnd:
Merged `std_extra` into `external`.

rikosellic (approved):

rikosellic:
Thanks for your contribution!

PR #431, opened 2026-05-02
Fix new mut ref

Marsman1996:
Since Verus have enabled the `new-mut-ref` by default in https://github.com/verus-lang/verus/pull/2393, we do not need to add `-V new-mut-ref` now.

And the token methods in VSTD are updated largely, the KVerus ports the old token code  back in the vstd_extra.
Maybe @rikosellic could have a better method?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ $(VERIFICATION_TARGETS):
 all: verify
 
 verify:
-	cargo dv verify --targets $(VERIFICATION_TARGETS) -- -V new-mut-ref
+	cargo dv verify --targets $(VERIFICATION_TARGETS)
 
 fmt:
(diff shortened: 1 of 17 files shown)


Marsman1996:
Of course CI would fail...

rikosellic:
The automatic synchronization of our Verus fork is blocked by an upstream CI change several days ago. So this PR should be postponed until I manually sync our Verus fork. I will try to update it today or tomorrow, but I can not guarantee I will have time to do this.

rikosellic:
Yeah, there is an exciting update of the resource algebra in `vstd`, I will rebase our definition onto it when I go back to work. Verus is suddenly making huge progress when we are on vacation. There will be some breaking changes.

PR #439, opened 2026-05-08
Bring segment.rs back into alignment with original

SNoAnd:
This PR focuses on eliminating transcription issues in `segment.rs`. It also refactors the permission structure of `SegmentOwner` to carry borrowed slot owners/perms instead of values, preventing duplication. We can do a similar refactor in linked_list later. The remaining work on `io.rs` will be next.

Github has the whole stack of commits listed again here, I guess because there was still some `phaseII/verified` history involved, but after this we'll stick to main.


rikosellic:
@SNoAnd  It seems this PR is not compatible with the latest Verus iterator update. Please refer to https://github.com/asterinas/vostd/pull/440  to fix it, thanks！

SNoAnd:
@rikosellic I believe it's working now.

Marsman1996:
@SNoAnd @hiroki-chen 
Does the DMA code modification right here?

hiroki-chen:
> @SNoAnd @hiroki-chen Does the DMA code modification right here?

The `dma` module is reverted again; needs fix.

ostd/specs/mm/vm_space_embedding.rs
hiroki-chen, line 1: Can you please elaborate more on the purpose of this `embedding` or give a concrete example? This appears confusing to me.

ostd/src/mm/dma/dma_stream.rs
hiroki-chen, line 1: This should be restored

hiroki-chen:
@SNoAnd, BTW Sean, the next time you open a new PR, you may want to rebase from main to avoid unrelated history commits.

ostd/specs/mm/vm_space_embedding.rs
SNoAnd: Yeah, this is just a WIP that I was playing with while waiting on a blocker from the `io` work, so take it with a grain of salt. I wanted to see how easy it would be to approximate a deep embedding in verus. A rigorous statement of soundness needs to do induction over arbitrary call sequences. Currently we make that argument on paper rather than in Verus, because Verus obviously can't quantify over functions. This file is an experiment in working around that limitation by defining each function identifier in a big enum and treating calls as transitions in a proof-mode state machine.

ostd/src/mm/dma/dma_stream.rs
SNoAnd: It's back, not sure why it disappeared again.

SNoAnd:
@hiroki-chen, yup! I should have done that this time. However, I am retiring the branch after this PR, which will have the same effect going forward.

rikosellic:
Thanks for your contribution！

Marsman1996:
@rikosellic 
The Copilot also does not identify the duplicate code in `vstd_extra/src/external/range.rs` and `vstd_extra/src/external/slice.rs`...

rikosellic:
I think Copilot only considers the patch, not the whole codespace.

ostd/src/mm/frame/segment.rs
```diff
@@ -8,16 +8,20 @@ use core::{fmt::Debug, ops::Range};
 use crate::mm::frame::{has_safe_slot, untyped::AnyUFrameMeta, Frame};
 use crate::mm::page_table::RCClone;
 
+use vstd_extra::assert;
 use vstd_extra::cast_ptr::*;
 use vstd_extra::cast_ptr::*;
```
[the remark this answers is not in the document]
hiroki-chen: You are absolutely right!

hiroki-chen:
Seems that copilot suggests that we lack error handling 🤔

Marsman1996:
> Seems that copilot suggests that we lack error handling 🤔

It somehow restore the original code of the Asterinas. 
For example, the original code of `VmReader<Fallible>::read_val` is 

```rust
    pub fn read_val<T: Pod>(&mut self) -> Result<T> {
        if self.remain() < core::mem::size_of::<T>() {
            return Err(Error::InvalidArgs);
        }

        let mut val = T::new_uninit();
        let mut writer = VmWriter::from(val.as_bytes_mut());
        self.read_fallible(&mut writer)
            .map_err(|(err, copied_len)| {
                // SAFETY: The `copied_len` is the number of bytes read so far.
                // So the `cursor` can be moved back to the original position.
                unsafe {
                    self.cursor = self.cursor.sub(copied_len);
                }
                err
            })?;
        Ok(val)
    }
```

SNoAnd:
Copilot makes a good point, the leaked permission would be an issue if we called the function from verified code. I was just working on that function already (there is a transcription error; the `Ok(_) => Ok(buf)` should be `Ok(buf) => Ok(buf)`) so I'll also make sure the permission gets rolled back.

PR #446, opened 2026-05-12
Add `default` interface for `VmSpace`

hiroki-chen:
Per the request in the doc, I added this as a member function as Verus forbids declaring preconditions for trait impls.

### DIFF
--- a/ostd/src/mm/vm_space.rs
+++ b/ostd/src/mm/vm_space.rs
@@ -147,6 +147,20 @@ type Result<A> = core::result::Result<A, Error>;
 
 #[verus_verify]
 impl<'a> VmSpace<'a> {
+    #[inline]
+    #[verus_spec(r =>
+        with Tracked(regions): Tracked<&mut MetaRegionOwners>,
+            Tracked(guards_k): Tracked<&mut Guards<'static, KernelPtConfig>>,
+            Tracked(guards_u): Tracked<&mut Guards<'static, UserPtConfig>>,
+        requires
+            old(regions).inv(),
+    )]
+    #[allow(private_interfaces)]
+    pub fn default() -> Self {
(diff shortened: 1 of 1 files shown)


Marsman1996:
So this is the convert of 
```rust
impl Default for VmSpace {
    fn default() -> Self {
        Self::new()
    }
}
```

@rikosellic  
Shall we add a comment of the original code to infer that there is a convertion?

hiroki-chen:
> So this is the convert of
> 
> ```rust
> impl Default for VmSpace {
>     fn default() -> Self {
>         Self::new()
>     }
> }
> ```
> 
> @rikosellic Shall we add a comment of the original code to infer that there is a convertion?

Well, there could be two ways:

- Convert unsupported traits into member functions but this requires some code changes.
- Use `ExTrait` and add a spec impl glue so that we can specify preconditions/postconditions but this might not be general for all cases and sometimes unstraightforward.

I chose to avoid the second one since sometimes Verus has a spec layer for `ExDefault` that forbids us to declare another one

Marsman1996:
> Well, there could be two ways:
> 
> * Convert unsupported traits into member functions but this requires some code changes.
> * Use `ExTrait` and add a spec impl glue so that we can specify preconditions/postconditions but this might not be general for all cases and sometimes unstraightforward.
> 
> I chose to avoid the second one since sometimes Verus has a spec layer for `ExDefault` that forbids us to declare another one

The conversion itself is not the problem. 

What I meant is that we may want to add a comment documenting the original code pattern before conversion. 
This would make future upstream synchronization or rebasing easier, because we can quickly understand what the verified version was converted from.

For example:
https://github.com/asterinas/vostd/blob/9c3832311ed0162b7022ea2d9ebebe351badffda/ostd/src/sync/rwlock.rs#L905-L913

hiroki-chen:
Surely it does. I can do a small PR to annotate the `mm` module.

PR #452, opened 2026-05-15
Io restoration

SNoAnd:
This PR restores parity with `io.rs` and fixes some incomplete verification issues in that file. I've streamlined the model a bit as well, and added part of the top-level soundness proof in `embedding`, mainly to help convince myself that the new model is sufficient.

The other interesting thing I'd like to point out is the tarball in `vtrace`. This is the static version of a code difference tracker we've been working on here: https://github.com/CertiKProject/dv-verus-rewrite-tracker. It documents all of the differences between this branch and the original `phaseII/mainline` code, subject to predefined equivalence relations. The idea is to expand the equivalences where possible, or else manually approve intentional differences. Segment and io have up-to-date triage in the static copy.


Marsman1996:
There are several extra files (i.e., `vtrace`?) uploaded, may need to remove.

SNoAnd:
Done! Also found a couple of redundant preconditions on the cursor and removed them.

ostd/libs/align_ext/src/lib.rs
rikosellic, line 1: I'm reconsidering whether this design is good for verification. When we call `align_up` in another function now, we will get the align condition for free in its post-condition, which means we will never discover an overflow bug in the future! Worse still, if we call this function on arguments that do not satisfy the condition, it will imply `false` and we can prove any statement.

rikosellic (changes requested):

ostd/libs/align_ext/src/lib.rs
rikosellic: Maybe we still need the panic condition in the precondition to detect incorrect calls to the verified function. We can do something like this:
```
fn exec_f()
requires
   !panic_condition()
ensures
   normal_post_condition(),
   panic_condition() ==> false,
```
  We can explicitly express the panic condition with `panic_condition() ==> false`, indicating that the function can never reach its end when the panic condition is met. ~And this condition is still inferred by your `vstd::assert!` macro. There may be more overflow checks enforced by Verus (e.g., `self + (align - 1) <= $uint_type::MAX` preserved in `align_up`'s precondition), we can insert more `vstd::assert!`s so that the `panic_condition` can cover those cases.~ (This is a direct contradiction) When we found that the panic condition is not fully described in the documentation, we should report it to the dev group.

rikosellic: @SNoAnd @hiroki-chen What do you think?

hiroki-chen: Well I think it's necessary to make any situation that may cause runtime panic *explicit* in the verification conditions just to avoid potential mistakes.

vstd_extra/src/panic.rs
```diff
@@ -12,4 +12,46 @@ pub fn panic_diverge() -> ! {
...
+/// and discharges `is_present()` (i.e. `self is Some` for `Option`) at
+/// the call site by branch elimination, so callers need not prove the
+/// value is populated.
+pub trait UnwrapOrPanic: Sized {
+    type Inner;
+
+    spec fn is_present(&self) -> bool;
+
+    spec fn payload(&self) -> Self::Inner;
+
```
hiroki-chen: Can we strengthen the postconditions to `self.is_present() <==> r == self.payload()`

hiroki-chen:
One thing about adding `axiom fn` for creating `tracked` objects to mimic their `spec` definitions: the current codebase uses `#[verifier::external_body] pub proof fn tracked_[fn_name]`. Can we please make the code style consistent @SNoAnd ?

ostd/libs/align_ext/src/lib.rs
rikosellic: The current design is equivalent to the explicit `requires true ensures panic_condition() ==> false`. But we can not rely on Verus to detect any misuse then.

SNoAnd: Reporting an unlabeled panic condition to the devs certainly makes sense. In this case the motivating caller was `write_vals`, which does document the panic. And shifting the panic condition back to the precondition and propagating it through to the API is a good way to identify whether the panic is documented.

The problem is that `requires !panic_condition()` doesn't actually guarantee that `panic_condition() ==> ⊥`. That was the issue we ran into with the overflows inside of the `assert!` macros: our Verus panic conditions were actually broader than the situations in which Rust panics, due to a mismatch in semantics between Verus spec addition and Rust addition.

`ensures !panic_condition()` (and `ensures panic_condition() ==> false`) is equivalent to `panic_condition() ==> ⊥` as far as Verus is concerned, but it holds trivially if we also include `requires !panic_condition()`.

SNoAnd: As far as concerns about proving everything go: one thing we could do is create our own `panics()` predicate with `#[verifier::opaque]` that unfolds to `false`. Then we would manually `reveal` it when we want to use the non-panic clause. Logically the same, but more intentional. It would also flag the postconditions that correspond to panics, so that we can more easily check them against the documentation.

rikosellic: > but it holds trivially if we also include requires !panic_condition()

Yeah, I also realized this shortly after my reply. So the current situation is somehow embarrassing: if we include the panic condition in the precondition, then our `vstd::assert` mechanism will have no effect. If we do not include it, then we can not propagate the panic condition to the call site automatically and may fail to identify undocumented panics. 

The addition semantic mismatch may be a smaller issue. We can add more constraints in the Verus condition to describe it more precisely.

vstd_extra/src/panic.rs
```diff
@@ -12,4 +12,46 @@ pub fn panic_diverge() -> ! {
     vpanic!("panic_diverge")
 }
 
+/// Extension trait providing a panicking `unwrap` that models Rust's
```
SNoAnd, line 23: @rikosellic, @hiroki-chen: This is another place where the panic discussion is relevant. Like with overflow arithmetic, Verus replaces the panic in `unwrap()` with a precondition, but ostd sometimes relies on the panic behavior.

rikosellic: This design is absolutely sound, as both we and the devs agree that panic diverge is a sound behaviour. The question is that not all panics are expected or preferred, which are also considered bugs (the fuzzing work by @Marsman1996 has found dozens of them). 

As a result, we may fail to identify that the following function may unexpectedly panic because Verus will no longer raise any verification errors:
```
pub fn api() {
       let x = None;
       x.unwrap_or_panic()
}
```
To prevent this, any function that is called in a verified caller should have the panic condition in its `requires` clause. For public APIs, we may keep the current design because they will not be called in a verification context in the near future.

ostd/libs/align_ext/src/lib.rs
SNoAnd: My worry is that the addition semantic mismatch might not be the only one. Turning panics and unexpected behavior into preconditions is a common Verus idiom.

I need to think harder about it, but maybe we can use the opaqueness idea on the `requires` side instead. The caller provides the non-panic predicate, but its content is only accessible once it's independently enforced by the `assert!`.

SNoAnd: I think I have a good solution. Panic diverge can take an `uninterp spec fn may_panic()` token in its requires. Functions that have the possibility of panicking must provide that token via their own requires. We write `!P || may_panic()` or `P ==> may_panic()` in the requires block, and propagate it back to the API surface.

This is framed as a "may" because the token on its own does nothing if the code doesn't actually reach a `panic_diverge`. If there is a single condition `P`, we can turn it into a total condition:

```rust
requires
    P <==> may_panic()
ensures
    !may_panic()
```

Working on implementation.

rikosellic: @SNoAnd Sounds like a great idea! 

To my understanding, do you mean something like `{may_panic} panic_diverge {false}`.

Then we have:
``
assert!(P)
``
is equivalent to:
``
{!P ==> may_panic()} if (P) {skip} else {panic_diverge()} {P}
``

I'm not very sure whether we should use `P<==> may_panic()` or just `P==>may_panic()` though.

SNoAnd: Yes, exactly! I think that `requires !P ==> may_panic()` and `ensures P` is actually sufficient, and much easier to work with. The leftward implication gets unpleasant when functions have multiple conditions that might trigger in different nested callees.

rikosellic:
@SNoAnd, please merge the latest main branch. We have enabled the global format check again!

vstd_extra/src/external/int_specs.rs
```diff
@@ -0,0 +1,34 @@
...
+use vstd::prelude::*;
+use vstd::arithmetic::power2::pow2;
+
+verus! {
+
+/// Spec for `<int_type>::is_power_of_two`. Returns true iff `self` is a positive
+/// power of two, i.e., there exists `e: nat` with `pow2(e) == self`.
+pub assume_specification[ u8::is_power_of_two ](self_: u8) -> (r: bool)
+    ensures
+        r <==> (exists|e: nat| pow2(e) == self_),
```
rikosellic, line 11: There are `is_pow2` and `is_pow2_exists` defined in `vstd::arithmetic::power2`, we can reuse those definitions.

rikosellic: We can also write a `macro_rule` to simplify this file.

rikosellic:
@SNoAnd Thanks for your contribution! Left a small comment in `vstd_extra`. Please also format the code (you may have to run `cargo dv bootstrap --upgrade` to get the latest version). I will merge it then.

rikosellic:
@hiroki-chen Is it valid to delete `read_byte` in `dma_stream`?

rikosellic (changes requested):

hiroki-chen:
> @hiroki-chen Is it valid to delete `read_byte` in `dma_stream`?

I think this is fine. The unverified `ostd` uses the default trait implementation and in the initial version I just marked this function as a placeholder.

SNoAnd:
> One thing about adding `axiom fn` for creating `tracked` objects to mimic their `spec` definitions: the current codebase uses `#[verifier::external_body] pub proof fn tracked_[fn_name]`. Can we please make the code style consistent @SNoAnd ?

Missed this in the shuffle, but yeah, we can do that.

rikosellic (approved):

rikosellic:
Thanks for your contribution! I think we can leave the minor improvements in another PR. I will merge it now.

PR #459, opened 2026-05-19
fix: breaking change from `git2` crate

Marsman1996:
Fix the breaking change caused by the `git2` crate from 0.20.3 -> 0.21.0.

Also upload the `Cargo.lock` file to avoid future breaking changes (unless yanked).


rikosellic:
Thanks for your contribution!

PR #460, opened 2026-05-19
Use `*mut` instead of `NonNull` in `NonNullPtr::ptr_perm_match``

rikosellic:
Minor adjustment to be compatible with rcu verification.


hiroki-chen:
hey @rikosellic, what's your progress right now on `rcu`? I also have some updates.

PR #463, opened 2026-05-19
Implement and enhance RCU with resource algebra proofs

hiroki-chen:


ostd/src/sync/rcu/mod.rs
```diff
@@ -101,37 +168,933 @@ pub struct RcuReadGuard<'a, P: NonNullPtr>(RcuReadGuardInner<'a, P>);
...
+            <P as NonNullPtr>::Target,
+            _,
+            RcuPtrGhost<P>,
+            _,
+        >,
     // We want to implement Send and Sync explicitly.
     // Having a pointer field prevents them from being implemented
     // automatically by the compiler.
-    _marker: PhantomData<*const P::Target>,
+    _marker: PhantomData<P>,
```
hiroki-chen: Will change it back to `_marker: PhantomData<*const <P as NonNullPtr>::Target>,`

rikosellic:
The correctness of the grace period recycling heavily depends on `DisablePreempt`. We should consider this once we take the more serious part of RCU. Not in the scope of this PR, though.

PR #471, opened 2026-05-22
Dma fix

hiroki-chen:


rikosellic (commented):
Is separating chaining function calls necessary?

hiroki-chen:
> Is separating chaining function calls necessary?

I'm not really sure about this. Perhaps this reduces the proof burden (?); you can try if inline function still verifies.

hiroki-chen:
> > Is separating chaining function calls necessary?
> 
> I'm not really sure about this. Perhaps this reduces the proof burden (?); you can try if inline function still verifies.

OK This is a regression.

ostd/src/sync/rwarc.rs
```diff
@@ -43,6 +43,46 @@ closed spec fn wf(&self) -> bool {
...
+    reveal(
+        <InvariantPredicate_auto_Inner_num_rw<_> as AtomicInvariantPredicate<
+            _,
+            usize,
+            int,
+        >>::atomic_inv,
+    );
+}
+
+fn inc_num_rw<T>(inner: &Inner<T>)
```
rikosellic: Is extracting this function necessary? It is not used at other places.

hiroki-chen:
The reason for the flaky proof is confusing. @rikosellic any ideas on the broken rwlock proof?

rikosellic:
That's strange, I will look into it today.

rikosellic:
~Turns out to be a `#[verus_spec]` bug, we may need a minimal example and report it.~

hiroki-chen:
Perhaps the verification resources have been diluted so that the assertion (though apparent) cannot be autoamtically deduced.

PR #473, opened 2026-05-22
Rename tracked versions of spec functions

SNoAnd:
Pretty simple, just renames `X_spec` to `X` and `Y` to `tracked_Y`.

### DIFF
--- a/ostd/specs/mm/frame/linked_list/linked_list_owners.rs
+++ b/ostd/specs/mm/frame/linked_list/linked_list_owners.rs
@@ -523,14 +523,12 @@ impl<M: AnyFrameMeta + Repr<MetaSlotSmall>> CursorOwner<M> {
         CursorOwner::<M> { list_own: list_own, index: index }
     }
 
-    #[verifier::returns(proof)]
-    pub axiom fn tracked_cursor_mut_at_owner(list_own: LinkedListOwner<M>, index: int) -> Self
-        returns Self::cursor_mut_at_owner(list_own, index);
+    pub axiom fn tracked_cursor_mut_at_owner(list_own: LinkedListOwner<M>, index: int) -> (tracked res: Self)
+        ensures res == Self::cursor_mut_at_owner(list_own, index);
 
-    #[verifier::returns(proof)]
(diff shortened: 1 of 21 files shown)


ostd/specs/mm/frame/linked_list/linked_list_owners.rs
```diff
@@ -510,30 +510,30 @@ impl<M: AnyFrameMeta + Repr<MetaSlotSmall>> CursorOwner<M> {
...
         CursorOwner::<M> { list_own: list_own, index: index }
     }
 
     #[verifier::returns(proof)]
-    pub axiom fn cursor_mut_at_owner(list_own: LinkedListOwner<M>, index: int) -> Self
-        returns Self::cursor_mut_at_owner_spec(list_own, index);
+    pub axiom fn tracked_cursor_mut_at_owner(list_own: LinkedListOwner<M>, index: int) -> Self
+        returns Self::cursor_mut_at_owner(list_own, index);
 
     #[verifier::returns(proof)]
```
rikosellic: @SNoAnd  Is `[verifier::returns(proof)]` an old-fashioned way to identify the function that returns a tracked value? Maybe we should use `(tracked res: Self)` now.

rikosellic (changes requested):

ostd/specs/mm/frame/linked_list/linked_list_owners.rs
hiroki-chen: @rikosellic BTW What's the most idiomatic way to defined trusted tracked proof functions

SNoAnd: @rikosellic, exactly. We can switch to the `tracked res` version.

rikosellic: > @rikosellic BTW What's the most idiomatic way to defined trusted tracked proof functions

I think we can just use axiom fn which returns tracked res.

rikosellic (approved):

rikosellic:
LGTM, thanks for your contribution！

PR #477, opened 2026-05-25
`PageTablePageMeta::on_drop` annotation

SNoAnd:
This PR is laying the groundwork to verify `PageTablePageMeta::on_drop` and has some fairly wide-reaching changes related to that. `AnyFrameMeta` is now dyn-compatible! I'm still not sure if we'll be able to verify most of the dyn functions, but this allows us to break the trait dependency cycle for `PageTablePageMeta::on_drop`. `AnyFrameMeta` gets a generic `OnDropArgs` type, since for this metadata we need a bunch of `verus_spec` parameters that we don't use for simpler kinds.

Next step is to verify the actual invariant preservation.

There is also some other precondition-trimming work from elsewhere in the codebase that hadn't gotten PR'd yet (the embedding commits).


rikosellic:
Good to see this progress！I'm on my way to work. Will review it very soon.

rikosellic:
Thanks for your contribution!

PR #480, opened 2026-05-26
chore: More idiomatic `VmIoOwner` definition

rikosellic:
To declare a ghost field in a `tracked struct`, use `ghost field: Type` instead of `field: Ghost<Type>`. @SNoAnd @hiroki-chen


hiroki-chen:
Thanks for the fix!

PR #486, opened 2026-06-01
Add `verified_bitflags` crate

rikosellic:
This is a simple Verus-verified version of the [`bitflags`](https://docs.rs/bitflags/latest/bitflags/) crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -547,12 +547,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "verified_bitflags"
+version = "0.1.0"
+dependencies = [
+ "vstd",
+]
+
+[[package]]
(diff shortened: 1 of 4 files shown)


hiroki-chen:
Would it be safer to introduce the "real" semantics of bitflags (i.e., use the `set` semantics)

rikosellic:
> Would it be safer to introduce the "real" semantics of bitflags (i.e., use the `set` semantics)

Yeah that is absolutly desired. Please feel free to creats PRs if you are availae. The current version is a bare minimum to support its usage in ostd. I'm now focusing on replacing those poorly defined flags and related functions in ostd.

hiroki-chen:
Sure, I will do it tomorrow.

PR #487, opened 2026-06-01
WIP: Weak memory support

hiroki-chen:
Ok so the weak memory has been started, and this serves as the triage PR for this feature. The core idea behind this implementation to encode message histories (parameterized by the `timerstamp` token) for each memory location as _ghost states_ in the concurrent separation logic which allows the caller to open the invariant and obtain the handle to the underlying data.

Basically, the roadmap includes:

- <s>[x] Modelling the "weak" part (i.e., memory ordering) of the atomic operations. This includes adding new atomic wrapper types from the Rust std library and stopping using the vstd's atomic types at all (since they assume SC). The Verus team is also designing this part.
- [x] Introduce idomatic macros, types, and interfaces for proving around our customized atomic types.</s>
- [x] Use the atomic types to prove the RCU reclamation as well as the traversal security properties.
- [x] Design the OS scheduler/task struct (for sound thread/preemption reasoning).
- [x] Add several examples.


hiroki-chen:
Reference https://dl.acm.org/doi/pdf/10.1145/3729246 and https://dl.acm.org/doi/pdf/10.1145/3591297

There is no need to model the full C11 semantics (they are even flawed) so we just do an in-order spec.

hiroki-chen:
Just done the low-level primitives so that we can model the atomic operations for the weak memory model.

rikosellic:
This will be a huge win if it is correct. I've not read the paper in detail yet, so I'm afraid I will not have enough knowledge to review this PR until next week.

hiroki-chen:
> This will be a huge win if it is correct. I've not read the paper in detail yet, so I'm afraid I will not have enough knowledge to review this PR until next week.

No worries; the core idea is rather simple. I'll keep you updated.

ostd/specs/sync/weak_memory.rs
```diff
@@ -0,0 +1,2010 @@
...
+        opens_invariants none
+        no_unwind
+    {
+        let result = self.value.compare_exchange(current, new, Ordering::AcqRel, Ordering::Acquire);
+        (result, Ghost::assume_new(), Tracked::assume_new())
+    }
+}
+
+#[cfg(verus_keep_ghost)]
+fn smoke_test_weak_atomic_with_ghost() {
```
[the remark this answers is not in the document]
hiroki-chen: Ignore this smoke test

hiroki-chen:
Now comes the hardest part. The current thing is that we cannot arbitrarily create a tracked view of the "current thread" so we need to obtain this from some authoritative state like `PreemptDisabledGuard`. But `disable_preempt` does not take any argument and the returned guard is minted out of thin air. On the one hand, this is meaningless for verification as everyone now can create an empty `ThreadView`; on the other hand, it is difficult to track a thread-local `view` across all call sites, I'm very unsure about this (`thread_local!` is a runtime trick).

hiroki-chen:
Verus cannot automatically "know" who is running now so this fact must be glued by the task/scheduler, but this requires the `exec` code to provide a tracked token from somewhere. My initial thought is to introduce a global invariant that keeps a map from thread id (as scheduler might know it) to their tracked information so that we can extract the information when scheduler picks a new task and that we can leverage this information to botain the ground-truth.

rikosellic:
> Verus cannot automatically "know" who is running now so this fact must be glued by the task/scheduler, but this requires the `exec` code to provide a tracked token from somewhere. My initial thought is to introduce a global invariant that keeps a map from thread id (as scheduler might know it) to their tracked information so that we can extract the information when scheduler picks a new task and that we can leverage this information to botain the ground-truth.

Is this map stored in the scheduler? For the thread ID token, we may use the [authoritative algebra](https://verus-lang.github.io/verus/verusdoc/vstd/resource/ghost_var/struct.GhostVarAuth.html), so one half of the token is stored in the scheduler, while the other is given to the thread.

hiroki-chen:
> > Verus cannot automatically "know" who is running now so this fact must be glued by the task/scheduler, but this requires the `exec` code to provide a tracked token from somewhere. My initial thought is to introduce a global invariant that keeps a map from thread id (as scheduler might know it) to their tracked information so that we can extract the information when scheduler picks a new task and that we can leverage this information to botain the ground-truth.
> 
> Is this map stored in the scheduler? For the thread ID token, we may use the [authoritative algebra](https://verus-lang.github.io/verus/verusdoc/vstd/resource/ghost_var/struct.GhostVarAuth.html), so one half of the token is stored in the scheduler, while the other is given to the thread.

Yes, this is a tracked struct in the scheduler. I think this may be a `LocalInvariant` or `AtomicInvariant` or something that stores the map.

hiroki-chen:
Ok, let me briefly explain what happend since the last change. To enable RCU-verification we started to implement models for the task thread views which are bound to the scheduler that acts as the authoritative resouce holder. The scheduler will keep track of the active threads as a finite `Map` from thread IDs to their corresponding resource tokens (thread views). At runtime we know which one is running by `current_task`so that we can open the invariant locally to obtain the token for the current thread if necessary.

This is a high-level idea and is subject to change.

rikosellic:
> runtime we know which one is running by `current_task`so that we can open the invariant locally to obtain the token for the current thread if necessary.

It sounds reasonable to me.

hiroki-chen:
Currently also blocked by the CPU core design.

hiroki-chen:
This PR would be very huge so better take a quick look for now. @rikosellic, @Marsman1996

rikosellic:
It will take some time, I'm afraid. BTW, Verus supports the Iris-style logical atomicity today. Will it bring some interesting interaction with the weak memory model?

hiroki-chen:
> It will take some time, I'm afraid. BTW, Verus supports the Iris-style logical atomicity today. Will it bring some interesting interaction with the weak memory model?

Definitely! I',m not very sure if this new design is just a syntactical sugar or some fundamental changes that faciliate the design. I'll take a look at this issue.

ostd/specs/sync/weak_memory.rs
```diff
@@ -0,0 +1,2593 @@
...
+use vstd::assert_sets_equal;
+use vstd::invariant::{AtomicInvariant, InvariantPredicate};
+use vstd::prelude::*;
+use vstd::resource::Loc;
+use vstd::resource::map::{GhostMapAuth, GhostPersistentPointsTo};
+use vstd::seq::Seq;
+
+verus! {
+
+// The "global" memory is defined wihtin the invariant we need to preserve and,
```
rikosellic: Typo`within`

ostd/specs/sync/weak_memory.rs
```diff
@@ -0,0 +1,2593 @@
...
+                        other.seen_at(id)
+                    } else {
+                        self.seen_at(id)
+                    },
+            ),
+        }
+    }
+
+    /// Partial ordering two threads' views.
+    pub open spec fn le(self, other: Self) -> bool {
```
rikosellic: You can define a `spec_le` so that you can use the `<=` operator.

ostd/specs/sync/weak_memory.rs
```diff
@@ -0,0 +1,2593 @@
...
+/// A thread-local weak-memory view.
+///
+/// `seen[id] = ts` means this thread has advanced past all messages for `id`
+/// older than `ts`; future reads from that atomic must not go backwards.
+///
+/// Typically, if another thread has published a message with timestamp `ts` for `id`,
+/// and the reader reads the message via some atomic operations, then the reader's
+/// thread view will advance to at least `ts` for `id`.
+pub ghost struct WmView {
+    pub seen: IMap<AtomicId, Timestamp>,
```
rikosellic: Shall we use `Map` instead of `IMap`?

hiroki-chen: Well I think this doesn't matter in our case. 🤔

rikosellic: Indeed

rikosellic:
I've skimmed through the first part of `weak_memory`, the definitions of `WmView`, `Msg` and `History` look good to me. I will read the paper and hopefully can give a more thorough review this week. BTW, many of the definitions can be moved to `vstd_extra`. They do not seem to be `ostd`-specific. You may add a `vstd_extra::atomic_weak` for example, mimicking `vstd::atomic_ghost`.

hiroki-chen:
> I've skimmed through the first part of `weak_memory`, the definitions of `WmView`, `Msg` and `History` look good to me. I will read the paper and hopefully can give a more thorough review this week. BTW, many of the definitions can be moved to `vstd_extra`. They do not seem to be `ostd`-specific. You may add a `vstd_extra::atomic_weak` for example, mimicking `vstd::atomic_ghost`.

Sounds good to me! I'll do it later.

ostd/specs/sync/weak_memory.rs
rikosellic: In the paper it is a finite map. Not sure whether it will bring any benefits.

hiroki-chen: Indeed, better align with the paper imo

hiroki-chen:
Note: somehow blocked by the CPU local/core design. This should be left as is until that part has been merged upstream.

rikosellic:
I notice there is also some work about the weak memory model on a Verus branch. https://github.com/verus-lang/verus/pull/2716

hiroki-chen:
> I notice there is also some work about the weak memory model on a Verus branch. [verus-lang/verus#2716](https://github.com/verus-lang/verus/pull/2716)

Seems that we all follow the same logic.

hiroki-chen:
This PR is large so in the future I'll split it into small pieces.

hiroki-chen:
Ready now!

hiroki-chen:
This one is done @rikosellic @Marsman1996

hiroki-chen:
Now this PR will be cherry-picked

Marsman1996:
Is there a roadmap or plan for Verus, about when to merge the irc11 branch into main?

hiroki-chen:
AFAIK, no clear plans for now

PR #490, opened 2026-06-02
More bitflags

hiroki-chen:
Just added the set semantics for this crate.


hiroki-chen:
OK It's broken; let me fix it.

PR #492, opened 2026-06-02
fix: trigger warnings and delete remaining old files

Marsman1996:
Add missing triggers, `unsafe` keyword, and fix for `bitflags` doc building error


verified_bitflags/src/lib.rs
```diff
@@ -90,7 +90,7 @@ macro_rules! bitflags {
             $vis struct $name {
                 /// The raw bits backing this flags value.
                 bits: $T,
-                flags: ::vstd::prelude::Ghost<::vstd::set::Set<[< __ghost $name >]>>,
+                ghost flags: ::vstd::prelude::Ghost<::vstd::set::Set<[< __ghost $name >]>>,
```
hiroki-chen: This is unnecessary. Either you use `ghost flags` or `::vstd::prelude::Ghost`; don't mix them.

Marsman1996:
@SNoAnd I noticed that the `unsafe` blocks were removed again in #485. Is that because the code has now been fully verified, so the `unsafe` blocks are no longer necessary?

verified_bitflags/src/lib.rs
```diff
@@ -105,7 +105,7 @@ macro_rules! bitflags {
                     }
                     Self {
                         bits: self.bits,
-                        flags: self.flags,
+                        flags: Ghost(Self::flags_from_bits(self.bits)),
```
hiroki-chen: Is this change necessary?

verified_bitflags/src/lib.rs
```diff
@@ -183,7 +183,13 @@ macro_rules! bitflags {
                         Self::from_bits_unchecked(bits),
                 {
                     use ::vstd::prelude::Ghost;
-                    Self { bits, flags: Ghost(Self::flags_from_bits(bits)) }
+                    Self {
```
hiroki-chen: What's the reason for this change

verified_bitflags/src/lib.rs
Marsman1996: You're right

verified_bitflags/src/lib.rs
Marsman1996: This is a workaround: since `flags_from_bits` error in `make doc` process, then we just replace it with `Ghost::assume_new()` during the doc building

rikosellic:
I plan to further improve `verified_bitflags` to make it an exact verified extension of the [`bitflags` crate](https://docs.rs/bitflags/latest/src/bitflags/lib.rs.html), with a set semantic. Please wait until then.

verified_bitflags/src/lib.rs
hiroki-chen: This is hacky but works. I cannot find any other workarounds

hiroki-chen (approved):
LGTM

rikosellic:
> I plan to further improve `verified_bitflags` to make it an exact verified extension of the [`bitflags` crate](https://docs.rs/bitflags/latest/src/bitflags/lib.rs.html), with a set semantic. Please wait until then.

This will take longer than expected. To avoid blocking the development, I've created a [PR](https://github.com/asterinas/vostd/pull/494) that slightly changes the implementation of `verified_bitflags`. Then we can merge this PR,

Marsman1996:
> > I plan to further improve `verified_bitflags` to make it an exact verified extension of the [`bitflags` crate](https://docs.rs/bitflags/latest/src/bitflags/lib.rs.html), with a set semantic. Please wait until then.
> 
> This will take longer than expected. To avoid blocking the development, I've created a [PR](https://github.com/asterinas/vostd/pull/494) that slightly changes the implementation of `verified_bitflags`. Then we can merge this PR,

Maybe I can revert the `bitflags` related commit and merge this PR first, then add the related logic in your PR

rikosellic:
I agree, this is the most efficient way.

verified_bitflags/src/lib.rs
rikosellic: Solved by https://github.com/asterinas/vostd/pull/494 because that field no longer exists.

rikosellic:
Actually it seems there is no conflicts any more.

rikosellic:
Thanks for your contribution!

PR #493, opened 2026-06-02
Add auxiliary traits to define spec-mode operators

rikosellic:
Verus automatically detects methods like `spec_add` to overload spec-mode `+` operators. However, the `vstd::prelude::SpecAdd` trait can not be implemented outside `vstd`. So this PR adds several simple traits so that we can define spec mode operators for our own types like this:
```
verus! {

trait SimpleSpecAdd {
    spec fn spec_add(self, other: Self) -> Self;
}

trait SimpleBitXor {
    spec fn spec_bitxor(self, other: Self) -> Self;
}

struct A {}

impl SimpleSpecAdd for A {
    spec fn spec_add(self, other: Self) -> Self {
        self
    }
}

impl SimpleBitXor for A {
    spec fn spec_bitxor(self, other: Self) -> Self {
        self
    }
}

fn main() {
    let a = A {};
    let b = A {};
    assert(a + b == a);
    assert(a ^ b == a);
}
```


hiroki-chen:
I think we define them as

```rs
#[verifier::external_trait_specification]
#[verifier::external_trait_extension(AddSpec via AddSpecImpl)]
pub trait ExAdd<T>: Sized {
    type ExternalTraitSpecificationFor: core::ops::Add<T>;

    spec fn obeys_add_spec() -> bool;

    spec fn spec_add(self, v: T) -> Self;

    fn add(self, v: Self) -> Self;
}
```

rikosellic:
If you mean [`vstd::std_specs::AddSpec`](https://verus-lang.github.io/verus/verusdoc/vstd/std_specs/ops/trait.AddSpec.html), here `add_spec` is purely a specification for `add` and is different from `spec_add`, which will not be automatically overloaded in spec mode operators.

hiroki-chen:
> If you mean [`vstd::std_specs::AddSpec`](https://verus-lang.github.io/verus/verusdoc/vstd/std_specs/ops/trait.AddSpec.html), here `add_spec` is purely a specification for `add` and is different from `spec_add`, which will not be automatically overloaded in spec mode operators.

That's awkward

rikosellic:
Yeah, it provides some strange flexibility that one can have different `add_spec` and `spec_add`, which brings more confusion than convenience. Maybe we can add an implementation of `SpecAdd` for all types with `AddSpec` so that we impose the equality in our project?

rikosellic:
@hiroki-chen What do you think about the following style?
```
verus! {

pub struct A {}

pub trait SpecAddTrait<Rhs = Self> {
    type Output;
    spec fn spec_add(self, rhs: Rhs) -> Self::Output;
}

impl<T: AddSpec<Rhs>, Rhs> SpecAddTrait<Rhs> for T {
    type Output = <T as Add<Rhs>>::Output;

    #[verifier::inline]
    open spec fn spec_add(self, rhs: Rhs) -> Self::Output {
        if Self::obeys_add_spec() { self.add_spec(rhs) } else { arbitrary() }
    }
}

impl Add for A {
    type Output = Self;

    fn add(self, rhs: Self) -> Self::Output {
        self
    }
}

impl AddSpecImpl for A {
    open spec fn obeys_add_spec() -> bool {
        true
    }

    open spec fn add_req(self, rhs: Self) -> bool {
        true
    }

    open spec fn add_spec(self, rhs: Self) -> Self {
        self
    }

}

fn main() {
    let a = A {};
    let b = A {};
    let c = a + b;
    assert(a + b == a);
    assert(a + b == c);
}
}
```

hiroki-chen:
> @hiroki-chen What do you think about the following style?
> 
> ```
> verus! {
> 
> pub struct A {}
> 
> pub trait SpecAddTrait<Rhs = Self> {
>     type Output;
>     spec fn spec_add(self, rhs: Rhs) -> Self::Output;
> }
> 
> impl<T: AddSpec<Rhs>, Rhs> SpecAddTrait<Rhs> for T {
>     type Output = <T as Add<Rhs>>::Output;
> 
>     #[verifier::inline]
>     open spec fn spec_add(self, rhs: Rhs) -> Self::Output {
>         if Self::obeys_add_spec() { self.add_spec(rhs) } else { arbitrary() }
>     }
> }
> 
> impl Add for A {
>     type Output = Self;
> 
>     fn add(self, rhs: Self) -> Self::Output {
>         self
>     }
> }
> 
> impl AddSpecImpl for A {
>     open spec fn obeys_add_spec() -> bool {
>         true
>     }
> 
>     open spec fn add_req(self, rhs: Self) -> bool {
>         true
>     }
> 
>     open spec fn add_spec(self, rhs: Self) -> Self {
>         self
>     }
> 
> }
> 
> fn main() {
>     let a = A {};
>     let b = A {};
>     let c = a + b;
>     assert(a + b == a);
>     assert(a + b == c);
> }
> }
> ```

If I understand correctly, this trait serves as the only proxy layer?

rikosellic:
I find that it will lead to multiple `spec_add` definitions for primitive types that have specialized spec mode behaviours (`usize`), so the automatic proxy will not work at all times. I will only add the traits then.

PR #494, opened 2026-06-02
verified_bitflags: Implement set semantic with `view` and remove ghost field

rikosellic:
@hiroki-chen It seems the set semantics can be implemented only using the view method, without a ghost field and a type invariant. Is this okay?


hiroki-chen:
It should be fine.

PR #498, opened 2026-06-03
fix: const value & visibility

Marsman1996:
Restore the const value of `FRAME_METADATA_CAP_VADDR` and `FRAME_METADATA_BASE_VADDR` to align with the 0.16.0. 
See https://github.com/asterinas/vostd/blob/f51a569a796b032fe362557a58bea69c633a6649/ostd/src/mm/kspace/mod.rs#L92-L95

I do not know why the current value is up to date with 0.17.0 (https://github.com/asterinas/asterinas/commit/6af524b4518694d6d8f6934b70daea5788a3a14f)

Besides that I also restore some of the visibility


rikosellic:
LGTM

PR #499, opened 2026-06-03
fix: `arch::mm::PageTableEntry` overhaul

rikosellic:
The code of `PageTableEntry` is neither at the correct place nor matches the source code.


rikosellic:
- Remove `specs/arch/x86/page_table_entry` that does not match the source code.
- Retire unverified `vstd_extra::bit_mapping`.
- Refine the properties in `PageTableEntryTrait`.
- Use and prove the correct source code in `src/arch/x86/mm`.

PR #501, opened 2026-06-04
Add `VmSpace::activate` back

hiroki-chen:
This API was deleted but for clearity I added it back.

### DIFF
--- a/ostd/src/mm/vm_space.rs
+++ b/ostd/src/mm/vm_space.rs
@@ -13,8 +13,6 @@ use vstd::prelude::*;
 use vstd::simple_pptr::PointsTo;
 use vstd::vpanic;
 
-use crate::specs::mm::virt_mem::{MemView, VirtPtr};
-
 use crate::error::Error;
 use crate::mm::frame::MetaSlot;
 use crate::mm::frame::untyped::UFrame;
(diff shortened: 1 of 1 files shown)


ostd/src/mm/vm_space.rs
```diff
@@ -313,6 +313,31 @@ impl<'a> VmSpace<'a> {
         }
     }
 
+    /// Activates the page table on the current CPU.
+    #[verifier::external_body]
+    pub fn activate(this: &RoArc<Self>) {
```
rikosellic, line 318: Why the function signature is `this: &RoArc` instead of `self: &Arc`

Marsman1996:
Well, later I will create a PR to add all deleted code back, in `/* */` comment.

ostd/src/mm/vm_space.rs
hiroki-chen: In case we can verify it later in the future I chose to use our `Arc` (there is no reference needed so `RwArc` should be sufficient). Also, `self: RwArc<Self>` will be rejected by Rust because this is ill-formed `self` so I changed its name to `this`

hiroki-chen:
> Well, later I will create a PR to add all deleted code back, in `/* */` comment.

For this I had an idea:

When anyone modifies the exec code they should declare a file like

```toml
[[anchor]]
git = "https://github.com/asterinas/asterinas.git"
tag = "0.16.0"
path = "ostd/src/mm/vm_space.rs"

[functions.verified]
name = "VmSpace::cursor_mut"
how = "transcribe"

[functions.unsurpported]
name = "VmSpace::activate"
reason = "lack-hardware-modelling"
```

and we can use LLM to design a tool to check if anything is left undeclared and if there is any mismatches in the function signature, declaration, etc.

PR #502, opened 2026-06-04
chore: add deleted logic back

Marsman1996:
No that big actually, it seems not much code is deleted


rikosellic:
That's a good news, thanks for your contribution!

PR #505, opened 2026-06-05
prove: mm/heap

Marsman1996:
Limited proof since mm/frame is not fully verified


ostd/src/mm/heap/mod.rs
```diff
@@ -17,6 +19,39 @@ pub use self::{
...
+}
+
+verus! {
+
+#[verifier::external_type_specification]
+#[verifier::external_body]
+pub struct ExLayout(Layout);
+
+#[verifier::external_type_specification]
+pub struct ExAllocError(AllocError);
```
rikosellic: We should put the external type specification in vstd_extra/external.

rikosellic (changes requested):

PR #509, opened 2026-06-08
Drop tracking and assorted cleanup fixes

SNoAnd:
The big change in this PR is an adjustment to how we keep track of dropped objects (and by extension, `ManuallyDrop`). The problem with the old approach was that it didn't keep track of when objects would normally be dropped by Rust, so while it modeled what happened with explicit `ManuallyDrop::new` calls, it didn't guarantee that we weren't missing such a call in a place that would have required it. This change instead handles frame drops uniformly: `MetaRegionOwners` tracks the extant frame objects that are in-scope somewhere in a live function, and will have drop called at them at some point in the future. By adding post-conditions that functions maintain the `frame_obligations` field unless they are intentionally returning a new frame object, we can catch places where an explicit call to `drop` is needed to model Rust's behavior.

Besides this, I've been working on the final report, and in the process finding small issues. Too-strong preconditions in `linked_list` and `segment`, mostly. Also I've been trying to make the requires and ensures blocks more readable by factoring out complex conditions into simpler functions. I expect to have a few more PRs in the future as I find and fix similar issues.


rikosellic:
Good to see this, `Drop` is a very important part for verification. Thanks for your contribution!

PR #511, opened 2026-06-10
Migrate to latest Verus; replace `=~=`

Marsman1996:
- [x] migrate to latest Verus to pass verification
- [x] fix the warnings


Marsman1996:
What the hell...
It works in my Mac...

ostd/specs/mm/page_table/cursor/owners.rs
```diff
@@ -927,7 +927,7 @@ impl<'rcu, C: PageTableConfig> CursorOwner<'rcu, C> {
         self.va.index.tracked_insert(self.level - 1, cont.idx as int);
         self.continuations.tracked_insert(self.level - 1, cont);
         assert(self.continuations == old(self).continuations.insert(self.level - 1, cont));
-        assert(self.va.index.dom() =~= Set::new(|i: int| 0 <= i < NR_LEVELS));
+        assert(self.va.index.dom() =~= Set::new_assuming_finite(|i: int| 0 <= i < NR_LEVELS));
```
rikosellic: [`new_assuming_finite`](https://verus-lang.github.io/verus/verusdoc/vstd/set/struct.Set.html#method.new_assuming_finite) is a deprecated function. The better way may be `Set::<int>::range`, accorfing to the [migration guide](https://verus-lang.github.io/verus/guide/spec_lib.html).

Marsman1996: Yes, fix the warning now...

rikosellic:
#499 has just been merged. Please keep the code updated.

ostd/specs/mm/page_table/owners.rs
```diff
@@ -1202,93 +1203,14 @@ impl<C: PageTableConfig> PageTableOwner<C> {
         }
     }
 
-    /// `view_rec` is finite: bounded by NR_ENTRIES branching and INC_LEVELS depth.
-    /// Proven by induction on `INC_LEVELS - path.len()`, collapsing the
-    /// `Set::new` existential into a `flatten` of a finite domain-map.
+    /// `view_rec` is finite (trivially true: `Set` is now always finite).
     pub proof fn view_rec_finite(self, path: TreePath<NR_ENTRIES>)
```
rikosellic: Do we still need this function anymore?

Marsman1996: nope

Marsman1996: will remove it, just test the linux verification result

rikosellic:
This PR makes me feel really good. We can get rid of a lot of annoying proofs.

verified_libs/vstd_extra/src/resource/ghost_resource/count.rs
```diff
@@ -100,12 +100,12 @@ impl<T, const TOTAL: u64> Protocol<(), T> for FractionalCarrierOpt<T, TOTAL> {
         }
     }
 
-    closed spec fn rel(self, s: Map<(), T>) -> bool {
+    closed spec fn rel(self, s: IMap<(), T>) -> bool {
```
rikosellic, line 103: ~~I'm wondering why we use IMap here, as it is obviously finite. Is it because Verus forgets () uin its finiteness support?~~

Gave it another thought and found that finiteness does not bring any benefits. It is okay.

hiroki-chen:
Which PR upgrades the Rust toolchain for CI?

Marsman1996:
> Which PR upgrades the Rust toolchain for CI?

This PR

hiroki-chen:
> > Which PR upgrades the Rust toolchain for CI?
> 
> This PR

A huge undertaking!

verified_libs/vstd_extra/src/resource/ghost_resource/count.rs
Marsman1996: Cause Verus update the signature for `rel`
https://github.com/verus-lang/verus/blob/e6a6d4fcc9601262a30e4eada28dcc0caedb01d4/source/vstd/resource/storage_protocol.rs#L44
```rust
/// See [`StorageResource`] for more information.
pub trait Protocol<K, V>: Sized {
    spec fn op(self, other: Self) -> Self;

    /// Note that `rel`, in contrast to [`PCM::valid`](crate::resource::algebra::ResourceAlgebra::valid), is not
    /// necessarily closed under inclusion.
    spec fn rel(self, s: IMap<K, V>) -> bool;
```

rikosellic: This makes a lot of sense, thanks!

Marsman1996:
There are 3 `new_assuming_finite` that KVerus fail to remove

hiroki-chen:
After this one is merged, we'll need some code refactoring.

rikosellic:
I think there is a more elegant way to solve the three `new_assuming_finite`s. I'm experimenting yet.

hiroki-chen:
`=~=` is default to `==` for set semantics, but it seems that KVerus still generated lots of unnecessary `=~=` for sets equivelances. @Marsman1996

ostd/specs/mm/page_table/cursor/cursor_steps.rs
```diff
@@ -2421,6 +2443,33 @@ impl<'rcu, C: PageTableConfig> CursorOwner<'rcu, C> {
 
             self.pop_level_owner_preserves_mappings();
             popped.move_forward_owner_preserves_mappings();
+        } else {
+            // level == NR_LEVELS, last index: move_forward_owner_spec only changes va/popped_too_high
+            let fwd = self.move_forward_owner_spec();
+            assert(fwd.continuations =~= self.continuations);
```
hiroki-chen: Should we put a hard constraint that the agents should never generate `=~=`s? This is quite annoying and creates inconsistencies across the codebase.

hiroki-chen (changes requested):

ostd/specs/mm/page_table/cursor/cursor_steps.rs
rikosellic: Yes, may be we should add it in the skills.

rikosellic: We can not replace `=~=` in `spec fn`s. But we can freely replace them in assertions and pre/post-conditions.

rikosellic: Correction: Verus promotes == to =~= inside assert, ensures, and invariant. I've replaced as many as possible. But we still need `=~=` in `requires` and `spec fn`s. We need to record it in a skill.

rikosellic (approved):

hiroki-chen (approved):

PR #515, opened 2026-06-12
fix missing tracked variable

hiroki-chen:
Moved the `const fn` block out of `verus_verify` and made tracked variable constructed inside the `Tracked` mode.

### DIFF
--- a/ostd/src/sync/rcu/mod.rs
+++ b/ostd/src/sync/rcu/mod.rs
@@ -232,7 +232,6 @@ impl<P: NonNullPtr> RcuInner<P> {
     }
 }
 
-#[verus_verify]
 impl<P: NonNullPtr + Send> RcuInner<P> {
     #[inline(always)]
     const fn new_none() -> (res: Self)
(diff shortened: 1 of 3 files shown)


rikosellic:
Please format the code.

rikosellic:
The CI using upstream verus is still failing, unfortunately.

ostd/src/sync/rcu/mod.rs
```diff
@@ -726,6 +727,16 @@ impl<'a, P: NonNullPtr + Send> RcuReadGuardInner<'a, P> {
...
 }
 
+impl<P: NonNullPtr + Send> Rcu<P> {
+    // Creates a new RCU primitive that contains nothing.
+    //
+    // This is a constant equivalence to [`RcuOption::new(None)`].
+    // #[inline(always)]
+    // pub const fn new_none() -> Self {
+    //     Self(RcuInner::new_none())
+    // }
```
hiroki-chen: @rikosellic Looking into this API this might be unsound for Asterinas to create an empty `Rcu` because we should have `RcuOption` which represents this semantic. I think for vanilla `Rcu`s we should never allow them to be nullable? How do you htink?

ostd/src/sync/rcu/mod.rs
```diff
@@ -786,14 +787,6 @@ impl<P: NonNullPtr + Send> RcuOption<P> {
         }
     }
 
-    /// Creates a new RCU primitive that contains nothing.
```
rikosellic, line 789: Why is this deleted? It is in the original code.

PR #531, opened 2026-06-16
prove: 12 `axiom` in page table and freshness

Marsman1996:

PR #532, opened 2026-06-16
fix: Decouple arch-specific constants from `PagingConstsTrait` trait definitions, fix specifications, and prove axioms.

rikosellic:
The current verification design maps hard-coded x86-specific constants to the `PagingConstsTrait` abstraction, which is incorrect because we can never extend the proof to other architectures. The new design is to prove the properties of abstractions (e.g., `CursorOwner` based on `PageTableConfig`) purely with the const declarations in `PagingConstTrait`. And only use the arch-specific constants (controlled by attributes) when proving definitions instantiated with `PagingConsts`.

### DIFF
--- a/ostd/specs/arch/x86/mod.rs
+++ b/ostd/specs/arch/x86/mod.rs
@@ -1,6 +1,7 @@
 use crate::mm::kspace::FRAME_METADATA_RANGE;
 use crate::mm::kspace::{LINEAR_MAPPING_BASE_VADDR, VMALLOC_BASE_VADDR, paddr_to_vaddr};
-use crate::mm::{Paddr, PagingConsts, Vaddr};
+use crate::mm::{Paddr, Vaddr, page_size, KERNEL_VADDR_RANGE};
+use crate::arch::mm::PagingConsts;
 use crate::specs::mm::frame::mapping::{
     META_SLOT_SIZE, lemma_meta_to_frame_soundness, meta_to_frame,
 };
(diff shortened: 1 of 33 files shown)


hiroki-chen:
please format

rikosellic:
And it turns out to be a big change, because `CursorOwner::inv` is actually defined with the x86-specific constants. We have something like 
```
&&& self.level <= 4 ==> {...}`
&&& self.level <= 3 ==> {....}
```
This is wrong because there are 3-level page tables in other architectures.

rikosellic:
This PR is currently suspended.

rikosellic:
Need to merge the latest `main`branch. Hopefully, it will not be too hard, as many changes originally proposed in this PR have been individually merged in #531 #534 #537.

Marsman1996:
Reduce from 215 newly added `assume`s to 25 `assume`s

ostd/specs/arch/x86/mod.rs
```diff
@@ -103,4 +108,254 @@ pub broadcast proof fn lemma_meta_frame_vaddr_properties(meta: Vaddr)
...
+    vstd_extra::external::ilog2::lemma_usize_ilog2_to32();
+    vstd::arithmetic::power2::lemma2_to64();
+    vstd::arithmetic::power2::lemma2_to64_rest();
+    vstd::bits::lemma_usize_pow2_no_overflow(48);
+}
+
+/// Proves the concrete values of `vaddr_make` for the x86_64 paging configuration.
+///
+/// For any `C: PagingConstsTrait`, since all configs share
+/// `BASE_PAGE_SIZE == 4096`, `nr_subpage_per_huge == 512`, and `NR_LEVELS == 4`:
```
rikosellic, line 135: This looks suspicious.  I do not think this assumption holds for all trait implementations.

ostd/src/mm/mod.rs
```diff
@@ -138,38 +136,52 @@ pub trait PagingConstsTrait: Clone + Debug + Send + Sync + 'static {
...
             0 < Self::BASE_PAGE_SIZE(),
             is_pow2(Self::BASE_PAGE_SIZE() as int),
-            Self::NR_LEVELS() > 0,
+            3 <= Self::NR_LEVELS() <= 4,
             is_pow2(Self::PTE_SIZE() as int),
             0 < Self::PTE_SIZE() <= Self::BASE_PAGE_SIZE(),
             // FIXME: remove this once we have a more general
             Self::BASE_PAGE_SIZE() == PAGE_SIZE,
             Self::NR_LEVELS() == NR_LEVELS,
             Self::BASE_PAGE_SIZE() / Self::PTE_SIZE() == NR_ENTRIES,
```
rikosellic, line 149: We should remove these hard-coded constant requirements here.

ostd/specs/arch/x86/mod.rs
rikosellic: I think LLM is misguided by the requirements in `PagingConstsTrait`.

rikosellic:
Superceded by #710.

PR #534, opened 2026-06-18
Fix unverified code and update cursor

hiroki-chen:


ostd/src/arch/x86/mm/mod.rs
```diff
@@ -105,6 +105,9 @@ impl PagingConstsTrait for PagingConsts {
     proof fn lemma_paging_consts_properties()
     {
         lemma_pow2_is_pow2_to64();
+        assert(PAGE_SIZE == 4096usize);
+        assert(NR_ENTRIES == 512usize);
+        assert(Self::BASE_PAGE_SIZE_spec() / Self::PTE_SIZE_spec() == NR_ENTRIES);
```
rikosellic: Are these proofs necessary? I recall there is a `lemma_nr_subpage_per_huge_eq_nr_entries`

ostd/src/mm/kspace/kvirt_area.rs
```diff
@@ -722,11 +754,16 @@ impl KVirtArea {
...
                 it.seq().len() == frames.len(),
+                0 <= mapped_pages <= frames.len(),
+                mapped_pages == it.index(),
                 cursor.0.barrier_va.start == range.start + map_offset,
                 cursor.0.barrier_va.end == range.end,
                 cursor.0.guard_level == NR_LEVELS as u8,
+                cursor.0.va <= cursor.0.barrier_va.end,
                 range.end - range.start == area_size,
                 cursor.0.va == range.start + map_offset + it.index() * PAGE_SIZE,
+                cursor.0.va as int == range.start as int + map_offset as int + mapped_pages
```
rikosellic: `as int` may be unnecessary

hiroki-chen: Looks like this is from Sean; will remove it.

PR #541, opened 2026-06-22
chore: styling and code alignment

hiroki-chen:


ostd/src/mm/kspace/kvirt_area.rs
```diff
@@ -91,18 +91,19 @@ pub struct RangeAllocator;
...
 
     #[verifier::external_body]
-    pub fn alloc(&self, size: usize) -> (r: Result<core::ops::Range<Vaddr>, RangeAllocError>)
+    #[verus_spec(res =>
         ensures
-            r == kvirt_alloc_spec(size),
-            r is Err <==> kvirt_alloc_oom_condition(size),
-    {
+            res == kvirt_alloc_spec(size),
+            res is Err <==> kvirt_alloc_oom_condition(size),
```
rikosellic: Here we can use ==

hiroki-chen: done

PR #543, opened 2026-06-22
`vstd_extra`: Remove `ModelOf`

rikosellic:
`ModelOf` seems like an unnecessary complication because it adds a transparent layer above `owner@`. This PR removes this trait. @SNoAnd, please take a look to decide whether this change is beneficial!


SNoAnd:
This makes sense to me.

SNoAnd (approved):

PR #546, opened 2026-06-23
New embedding ops, axiom cleanup, `in_scope` removal

SNoAnd:
This PR contains a few different threads.
- Before adding the drop obligation system, `EntryOwner`'s `in_scope` field distinguished an owner that corresponded to a live `Entry` object from a parked `E` value for purposes of the reference count conditions. This was fragile, and is unnecessary with drop obligations handling the reference count arithmetic. Removing that is the most important part of this PR.
- I've added `kspace`, `segment`, `unique`, and `linked_list` to `mm/specs/embedding`. The embedding tracks the reference counting arithmetic across sequential calls, which is an important check on the individual function specs.
- Proved `lemma_view_in_vaddr_range`, previously an axiom, which was one of the per-config lemmas that got axiomatized because it depended on different configs behaviors. I also deleted a few axioms that weren't being used, no sense in cluttering the trust boundary any more than it already is.

### DIFF
--- a/ostd/specs/mm/embedding/kvirt_store.rs
+++ b/ostd/specs/mm/embedding/kvirt_store.rs
@@ -1,0 +1,198 @@
+//! Kernel-side one-step-soundness store for the kernel virtual area API
+//! ([`crate::mm::kspace::kvirt_area::KVirtArea`]) — the *kernel* analog of
+//! [`super::VmStore`].
+//!
+//! # Why a separate, kernel store
+//!
+//! [`super::VmStore`] models a caller of the **user** `VmSpace` API: it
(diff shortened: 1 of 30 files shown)


rikosellic (changes requested):
Please resolve the conflicts, as we have made some minor changes on the main branch.

BTW, are there any upcoming PRs in this phase that will make major changes to the verification design?

SNoAnd:
Conflicts are resolved. I have no plans to make any further architectural changes. I might make some changes to `embedding`, and there are the documentation enhancements you requested.

rikosellic (approved):
Thanks for your contribution!

PR #550, opened 2026-06-23
add iterator for segment

hiroki-chen:


hiroki-chen:
Finally we can support customized iterators but if we need to take extra arguments we'll have to maintain a dedicated iterator rather than simply implementing `Iterator` for `Segment`.

ostd/src/mm/frame/segment.rs
```diff
@@ -859,138 +862,325 @@ impl<M: AnyFrameMeta + Repr<MetaSlotStorage> + OwnerOf> Segment<M> {
         range
     }
 
-    /// Gets the next frame in the segment.
```
Marsman1996, line 862: Why the comment is deleted

Marsman1996 (changes requested):

ostd/src/mm/frame/segment.rs
hiroki-chen: Will add it back

rikosellic:
Well, I still prefer directly including the prophecy field in `Segment`. But I agree to merge this PR first, as the code semantics have not changed. I will experiment with the other design later.

ostd/src/mm/frame/segment.rs
```diff
@@ -859,137 +862,321 @@ impl<M: AnyFrameMeta + Repr<MetaSlotStorage> + OwnerOf> Segment<M> {
...
-    ///
-    /// This is the verified counterpart of [`Iterator::next`] for `Segment<M>`.
```
Marsman1996, line 864: Actually my concern is these properties-related comments, which are currently shown in our [doc](https://asterinas.github.io/vostd/ostd/mm/frame/segment/struct.Segment.html#verified-properties-3)

hiroki-chen: Ok! I will try to add it back

rikosellic (approved):
LGTM. Thanks for your contribution!

Marsman1996 (approved):
LGTM

PR #551, opened 2026-06-24
Adapt page table cursor proofs for macOS verification

Je5s1e:
## Summary

This PR adapts the page table cursor proofs so the `page_table` verification path can pass on macOS, where the previous larger proof obligation could fail during `make`.

The main changes are:

- Split part of the `take_next` frame-unmap metaregion preservation reasoning into a dedicated proof helper, `take_next_remove_path_preserves_metaregion`.
- Make several `map` proof obligations more explicit about metadata slot existence before indexing `slot_owners`.
- Keep the change proof-oriented and localized to the page table cursor proof path.

This does not intend to change runtime behavior.

 ## Verification
- `make`


rikosellic:

Marsman1996:
Maybe a CI for macOS is needed

PR #552, opened 2026-06-24
fix: remove auxiliary `MetaSlot::get_from_in_use_loop`; remove some re-exports

rikosellic:


rikosellic:
The original proof of `get_from_in_use` is very verbose. I have deleted 200 lines of proofs, leaving fewer than 30.

PR #556, opened 2026-06-25
ci: add check for verification warnings

Marsman1996:
Due to the large number of `#[verus_spec] is likely used inside a verus! block` warnings, dev might ignore the missing trigger warnings.

With this PR, the CI will report error for missing trigger

See https://github.com/Marsman1996/vostd/actions/runs/28144455441


Marsman1996:
Also add detection for `vstd::set::Set::<A>::finite` and `vstd::set::Set::<A>::new_assuming_finite`

PR #557, opened 2026-06-25
resolve CI verification warnings

Je5s1e:
## Summary

Fix three categories of verification warnings that cause the `Check for verification warnings` CI step to fail.

### 1. Remove deprecated `Set::finite()` (5 occurrences)
All `Set`s are finite since the Verus migration — the method is always `true` and redundant in `inv()` specs, `requires`, and `assert` statements.

### 2. Suppress `Set::new_assuming_finite()` deprecation (4 occurrences)
Replacing with `Set::new(f).unwrap()` would require proving ISet finiteness, which is non-trivial for these spec-only `used` sets. Added `#[allow(deprecated)]` to the four proof functions as a temporary suppression, consistent with the deprecation note's guidance for incremental migration.

### 3. Add explicit `#[trigger]` annotations (2 occurrences)
Verus had low confidence in auto-chosen triggers for two quantifiers in `owners.rs`. Added `#[trigger]` to silence the `note: automatically chose triggers for this expression` messages.

## Files Changed
- `ostd/specs/mm/embedding/kvirt_store.rs` — remove `.finite()`
- `ostd/specs/mm/embedding/list_store.rs` — remove `.finite()`, add `#[allow(deprecated)]`
- `ostd/specs/mm/embedding/mod.rs` — remove `.finite()` (3 places)
- `ostd/specs/mm/page_table/cursor/owners.rs` — add `#[trigger]`

### DIFF
--- a/ostd/specs/mm/embedding/list_store.rs
+++ b/ostd/specs/mm/embedding/list_store.rs
@@ -1069,7 +1069,6 @@ impl<M: AnyFrameMeta + Repr<MetaSlotSmall>> ListStore<M> {
     /// `LinkedList::push_front`: move the loose handle `lid` to the front
     /// of list `id`. The frame is forgotten into the list (its
     /// drop-obligation consumed); the `loose` entry is removed.
-    #[allow(deprecated)]
     pub proof fn step_push_front(tracked &mut self, id: ListId, lid: LooseId)
         requires
             old(self).inv(),
@@ -1082,7 +1081,7 @@ impl<M: AnyFrameMeta + Repr<MetaSlotSmall>> ListStore<M> {
         let ghost old_regions = self.regions;
         let ghost fidx = self.loose[lid].slot_index;
         // The other lists' ids — the lazily-minted id must avoid these.
-        let ghost used = Set::new_assuming_finite(
+        let ghost used = Set::<u64>::full().unwrap().filter(
             |x: u64|
                 (exists|i: ListId| #[trigger]
                     old_self.lists.dom().contains(i) && i != id && old_self.lists[i].list_id == x)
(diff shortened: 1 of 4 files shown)


ostd/specs/mm/embedding/list_store.rs
```diff
@@ -1374,6 +1372,7 @@ impl<M: AnyFrameMeta + Repr<MetaSlotSmall>> ListStore<M> {
     /// `LinkedList::push_back`: move the loose handle `lid` to the back
     /// of list `id`. Same global effect as [`Self::step_push_front`] —
     /// only the link's position within the list differs.
+    #[allow(deprecated)]
```
Marsman1996: We shall not use `#[allow(deprecated)]` here, it only bypasses the warning instead of solving it

ostd/specs/mm/embedding/list_store.rs
```diff
@@ -1659,6 +1658,7 @@ impl<M: AnyFrameMeta + Repr<MetaSlotSmall>> ListStore<M> {
     /// loose handle `lid` into list `id` at index `n` (`0 <= n <= len`).
     /// The general form of [`Self::step_push_front`] /
     /// [`Self::step_push_back`]; same global effect.
+    #[allow(deprecated)]
```
Marsman1996: Same here

Marsman1996 (changes requested):

rikosellic:
Please refer to [this guide](https://verus-lang.github.io/verus/guide/spec_lib.html#constructing-and-using-seq-set-iset-map-and-imap) to resolve the `new_assuming_finite` warnings. I think `Set::<u64>::full().unwrap()` will do in our use cases.

rikosellic:
BTW, Category I is already covered in #552.

Marsman1996:
？

Marsman1996 (approved):
LGTM, let's see whether it need a merge or not

rikosellic:

rikosellic (approved):

PR #560, opened 2026-06-25
prove: narrow `item_into_raw_spec_level_bounds` axiom to proof fn

Je5s1e:
Convert the bodyless `axiom fn item_into_raw_spec_level_bounds` into a checked `proof fn`.


ostd/src/mm/kspace/mod.rs
```diff
@@ -406,10 +406,29 @@ unsafe impl PageTableConfig for KernelPtConfig {
 
 impl KernelPtConfig {
     /// The spec agrees with the exec, which ensures 1 <= level <= NR_LEVELS.
-    pub axiom fn item_into_raw_spec_level_bounds(item: MappedItem)
+    /// The Tracked branch is proved; the Untracked branch needs `1 <= level <= NR_LEVELS`
+    /// as a precondition (moved from the axiom to caller obligations).
```
Marsman1996: @rikosellic I think this comment is unnecessary

Marsman1996 (changes requested):

ostd/src/mm/kspace/mod.rs
rikosellic: Indeed

ostd/src/mm/kspace/mod.rs
```diff
@@ -406,10 +406,29 @@ unsafe impl PageTableConfig for KernelPtConfig {
...
+    pub proof fn item_into_raw_spec_level_bounds(item: MappedItem)
+        requires
+            match item {
+                MappedItem::Untracked(_, level, _) => 1 <= level <= NR_LEVELS,
+                _ => true,
+            },
```
rikosellic: This clause can be written in a more idiomatic way:
`item matches MappedItem::Untracked(_,level,_)==> 1 <= level <= NR_LEVELS`

rikosellic (approved):
LGTM， thanks for your contributions!

Marsman1996 (approved):
LGTM

PR #568, opened 2026-06-26
Proof cleanup

rikosellic:
Powered by Deepseek-v4-pro.


rikosellic:
I believe there are many more redundant proofs.

PR #561, opened 2026-06-25
Documentation updates, small embedding update

SNoAnd:
This PR contains some doc updates meant to go along with the updated blog post. Also a fix for a behavioral mismatch in `vm_store` in the embedding.

PR #569, opened 2026-06-28
Documentation updates, small embedding update

SNoAnd:
This PR contains some doc updates meant to go along with the updated blog post. Also a fix for a behavioral mismatch in vm_store in the embedding. This is the same content as #561, but rebased instead of merged, so we shouldn't have any regressions.


rikosellic:
LGTM! Thanks for your contribution!

PR #570, opened 2026-06-29
prove axiom fn in tlb.rs

Je5s1e:
This PR removes two axiom-backed tracked helpers from `ostd/specs/mm/tlb.rs` by providing concrete proof implementations for them.


rikosellic:
This looks a bit strange. The `TlbModel` is a `tracked` struct while all fields are `ghost`. I believe this is a verification design issue. I will look into it later.

ostd/specs/mm/tlb.rs
```diff
@@ -10,9 +10,9 @@ use vstd_extra::ownership::*;
 
 verus! {
 
-pub ghost struct TlbModel {
-    pub pending: Seq<TlbFlushOp>,
-    pub mappings: Set<Mapping>,
+pub tracked struct TlbModel {
+    pub ghost pending: Seq<TlbFlushOp>,
+    pub ghost mappings: Set<Mapping>,
```
hiroki-chen, line 14: This looks weird to me, why do you want to keep `ghost` inside `tracked`?

ostd/specs/mm/tlb.rs
```diff
@@ -39,27 +39,33 @@ impl TlbModel {
         TlbModel { pending: self.pending, mappings: self.mappings.insert(m) }
     }
 
-    pub axiom fn tracked_update(&mut self, pt: PageTableView, va: Vaddr)
+    pub proof fn tracked_update(tracked &mut self, pt: PageTableView, va: Vaddr)
```
hiroki-chen, line 41: The idiomatic way for doing so is to define a pure function, e.g., `pub open spec fn update()` and then define `pub axiom fn tracked_update()` to lift the spec to tracked mode.

Because this `proof fn` does not prove anything (you just want to do some tracked operations), there is no need to define the body.

hiroki-chen (changes requested):

hiroki-chen:
> This looks a bit strange. The `TlbModel` is a `tracked` struct while all fields are `ghost`. I believe this is a verification design issue. I will look into it later.

I also left some comments.

PR #572, opened 2026-06-29
fix: rlimit in mm::kspace::kvirt_area

Marsman1996:
by adding checks for slot owners in KVirtArea frame mapping


rikosellic:
This is a neat fix, thanks!

PR #577, opened 2026-06-30
fix: simplification in frame

rikosellic:


ostd/specs/mm/frame/segment.rs
```diff
@@ -104,6 +104,7 @@ pub open spec fn seg_nframes(range: Range<Paddr>) -> int {
     (range.end - range.start) / PAGE_SIZE as int
 }
 
+// FIXME: is this necessary?
```
hiroki-chen, line 107: Do you mean it can be included as part of `Segment`?

rikosellic: I think we can just delete `SegmentOwner`, it does not seem to provide any information,

hiroki-chen: You are right. Sean re-designed this struct; previously this keeps track of the frame permission but now it semss redundant.

ostd/src/mm/frame/segment.rs
```diff
@@ -888,6 +865,7 @@ impl<M: AnyFrameMeta + Repr<MetaSlotStorage> + OwnerOf> Segment<M> {
 pub type SegmentIteratorItem<M> = (Frame<M>, Tracked<DropObligation<usize>>);
 
 /// Prophetic sequence state used to specify the frames that a segment iterator will yield.
+/// FIXME: Do we need another iterator struct?
```
hiroki-chen, line 868: Well, this would be neat to have one.

PR #578, opened 2026-06-30
directly return `&mut M` for `meta_mut`

hiroki-chen:


hiroki-chen:
#579

rikosellic:
Thanks for your contribution!

PR #582, opened 2026-06-30
prove two axiom fn in ostd/specs/mm/vm_space_embedding.rs

Je5s1e:


rikosellic:

PR #591, opened 2026-07-01
fix: error of cargo verus

### LINKED ISSUE #585: `cargo verus` is buggy
I've tried and found that `cargo verus verify` fails on our project. It seems the `pub use crate` in `verified_bitflags` is not correctly exported. It is a `rust_verify` bug. We should create a minimal example and report this bug to the Verus team, but it is rather bothersome.

Marsman1996:
fix #585


Marsman1996:
tomorrow I will test the new simplified skill for the second commit

rikosellic:
This is good news! We can migrate our infrastructure to `cargo verus`.

PR #593, opened 2026-07-02
prove: align bitflags to upstream

Je5s1e:
This PR updates `verified_bitflags` to better align its core raw-bit behavior with the upstream [`bitflags`](https://docs.rs/bitflags/latest/bitflags/) crate.

It adjusts `contains`, `intersects`, known/unknown bits handling, truncation, complement, and zero-bit flag behavior to follow raw bit semantics.


verified_libs/verified_bitflags/src/lib.rs
```diff
@@ -315,126 +317,98 @@ macro_rules! bitflags {
...
+                /// The bits in `self` that correspond to declared flags.
+                $vis closed spec fn known_bits_spec(&self) -> $T {
+                    self.bits() & Self::all_bits()
+                }
+
+                #[verifier::when_used_as_spec(known_bits_spec)]
+                $vis const fn known_bits(&self) -> (r: $T)
+                    returns self.known_bits(),
+                {
+                    self.bits & Self::all_bits()
```
rikosellic: In the [source code](https://github.com/bitflags/bitflags/blob/9253889ee93921055309908c46272bd6c5c6ad99/src/traits.rs#L168), there is no `all_bits()`, only `all().bits()`, can you adjust the code accordingly?

rikosellic (changes requested):

rikosellic:
BTW, Bits is a [trait](https://github.com/bitflags/bitflags/blob/9253889ee93921055309908c46272bd6c5c6ad99/src/traits.rs#L376) in the source code. Can you verify this in Verus? You may try this in another PR.

Marsman1996:
Actually I'm keeping thinking that since `verified_bitflags` is inside `‎verified_libs`, maybe we can rename it back to `bitflags`? @rikosellic

rikosellic:
Indeed, I think it is a good change.

Marsman1996:
> ... rename it back to `bitflags`

@Je5s1e

verified_libs/bitflags/src/lib.rs
```diff
@@ -579,6 +582,19 @@ macro_rules! bitflags {
...
+                    Self::from_bits_unchecked_spec(!self.bits() & Self::all().bits())
+                }
+
+                #[verifier::when_used_as_spec(complement_spec)]
+                $vis const fn complement(self) -> (r: Self)
+                    ensures
+                        r.bits() == (!self.bits() & Self::all().bits()),
+                    returns self.complement(),
+                {
+                    Self::from_bits_unchecked(!self.bits & Self::all().bits())
```
rikosellic: The source code of `complement` is not consistent with the [source code](https://github.com/bitflags/bitflags/blob/9253889ee93921055309908c46272bd6c5c6ad99/src/traits.rs#L368).

rikosellic (changes requested):

verified_libs/bitflags/src/lib.rs
```diff
@@ -473,15 +464,15 @@ macro_rules! bitflags {
...
+                        r is Some == ((bits & Self::all().bits()) == bits),
                         r matches Some(flags_value) ==> {
                             &&& flags_value.bits() == bits
                             &&& flags_value.flags_spec() == Self::flags_from_bits(bits)
                         },
                     returns
                         Self::from_bits(bits),
                 {
-                    if (bits & Self::all_bits()) == bits {
+                    if (bits & Self::all().bits()) == bits {
```
rikosellic: Please also fix this one.

rikosellic:
Can you also add these two [lemmas](https://github.com/asterinas/vostd/blob/a7a0d0fce4365dc7deed319b925a467cf0bf058e/ostd/src/mm/page_prop.rs#L156-L170) in `bitflags`?

rikosellic (approved):

PR #607, opened 2026-07-06
simplify: mm::io

Marsman1996:
Continue the work in #591, simplified with the KVerus simplify skill


rikosellic:
LGTM. We should eliminate these verbose proof codes.

PR #612, opened 2026-07-06
Prove page table item roundtrip axioms

Je5s1e:


ostd/src/mm/page_table/mod.rs
```diff
@@ -324,12 +324,11 @@ pub unsafe trait PageTableConfig: Clone + Debug + Send + Sync + 'static {
...
+        requires
+            has_safe_slot(paddr),
+            Self::item_from_raw_spec(paddr, level, prop) == item,
         ensures
-            Self::item_into_raw_spec(item) == (paddr, level, prop) <==> Self::item_from_raw_spec(
-                paddr,
-                level,
-                prop,
-            ) == item,
+            Self::item_into_raw_spec(item) == (paddr, level, prop),
```
rikosellic, line 342: This new definition only contains one direction, and it is not the direction documented in the source code: 
```
if the provided raw form matches the item that was consumed by
///    `item_into_raw`, `item_from_raw` restores the exact item that was
///    consumed by `item_into_raw`.
```
Can the definition be the opposite?

rikosellic (changes requested):

ostd/src/mm/page_table/mod.rs
```diff
@@ -324,6 +324,9 @@ pub unsafe trait PageTableConfig: Clone + Debug + Send + Sync + 'static {
     ;
 
     proof fn item_roundtrip(item: Self::Item, paddr: Paddr, level: PagingLevel, prop: PageProperty)
+        requires
+            has_safe_slot(paddr),
+            Self::item_from_raw_spec(paddr, level, prop) == item,
```
rikosellic: The second requirement makes the `item` parameter redundant; you may delete this parameter if you find it is unnecessary later.

rikosellic (changes requested):
It is still not correct. As you add `Self::item_from_raw_spec(paddr,level,prop) == item`, the post-condition is actually equivalent to `Self::item_into_raw_spec(item) == (paddr, level, prop)`. But I want something like `Self::item_into_raw(item) == (paddr,level,prop) ==> Self::item_from_raw(paddr, level. prop) == item`.

ostd/src/mm/page_table/mod.rs
```diff
@@ -323,15 +323,32 @@ pub unsafe trait PageTableConfig: Clone + Debug + Send + Sync + 'static {
             !Self::tracked(item) ==> new_regions.frame_obligations == old_regions.frame_obligations,
     ;
 
+    /// If the provided raw form matches an item consumed by `item_into_raw`,
+    /// then `item_from_raw` restores that exact item.
     proof fn item_roundtrip(item: Self::Item, paddr: Paddr, level: PagingLevel, prop: PageProperty)
```
rikosellic: Do we still need this lemma?

Je5s1e: item_roundtrip is no longer used now, I will remove this lemma

ostd/src/mm/vm_space.rs
```diff
@@ -1671,10 +1674,8 @@ unsafe impl PageTableConfig for UserPtConfig {
 
         Self::item_from_raw_spec_frame_ptr(pa, level, prop);
         let item = Self::item_from_raw_spec(pa, level, prop);
```
rikosellic, line 1684: This line is redundant now.

ostd/src/mm/page_table/mod.rs
```diff
@@ -323,13 +323,17 @@ pub unsafe trait PageTableConfig: Clone + Debug + Send + Sync + 'static {
...
 
-    proof fn item_roundtrip(item: Self::Item, paddr: Paddr, level: PagingLevel, prop: PageProperty)
+    /// The item reconstructed from a page-table entry is canonical for that
+    /// entry's raw form.
+    proof fn item_from_raw_roundtrip(paddr: Paddr, level: PagingLevel, prop: PageProperty)
+        requires
+            has_safe_slot(paddr),
         ensures
-            Self::item_into_raw_spec(item) == (paddr, level, prop) <==> Self::item_from_raw_spec(
+            Self::item_into_raw_spec(Self::item_from_raw_spec(paddr, level, prop)) == (
```
rikosellic: It should be `item_from_raw(item_into_raw(item)) == item`

ostd/src/mm/page_table/mod.rs
```diff
@@ -323,13 +323,34 @@ pub unsafe trait PageTableConfig: Clone + Debug + Send + Sync + 'static {
...
+            Self::item_into_raw_spec(item) == (paddr, level, prop),
+            Self::tracked(Self::item_from_raw_spec(paddr, level, prop)) == Self::tracked(item),
         ensures
-            Self::item_into_raw_spec(item) == (paddr, level, prop) <==> Self::item_from_raw_spec(
+            Self::item_from_raw_spec(paddr, level, prop) == item,
+    ;
+
+    /// The item reconstructed from a page-table entry is canonical for that
+    /// entry's raw form.
+    proof fn item_into_raw_roundtrip(paddr: Paddr, level: PagingLevel, prop: PageProperty)
```
rikosellic: Is it possible to delete this lemma?

ostd/src/mm/page_table/mod.rs
```diff
@@ -323,13 +323,34 @@ pub unsafe trait PageTableConfig: Clone + Debug + Send + Sync + 'static {
...
+        item: Self::Item,
+        paddr: Paddr,
+        level: PagingLevel,
+        prop: PageProperty,
+    )
+        requires
+            has_safe_slot(paddr),
+            Self::item_well_formed(item),
+            Self::item_into_raw_spec(item) == (paddr, level, prop),
+            Self::tracked(Self::item_from_raw_spec(paddr, level, prop)) == Self::tracked(item),
```
rikosellic: Can this tracked condition be moved into the postcondition? Or just delete it.

rikosellic:
Generally looks good to me. Please resolve the two comments.

rikosellic:
I mean deleting `item_into_raw_roundtrip`.

Marsman1996:
@rikosellic 

It seems to remove the `Self::tracked(Self::item_from_raw_spec(paddr, level, prop)) == Self::tracked(item),`, we have to add new post-condition for `axiom fn item_into_raw_spec_untracked` and `axiom fn item_into_raw_spec_tracked_prop`

And `Self::tracked(Self::item_from_raw_spec(paddr, level, prop)) == Self::tracked(item),` itself is a weak version of `Self::item_from_raw_spec(paddr, level, prop) == item,`, so add it to post-condition is rebundant

rikosellic:
I'm not sure about the tracked thing, and I will check it later. The roundtrip lemma looks good. I will merge it; thanks for your contribution!

rikosellic (approved):

PR #613, opened 2026-07-06
fix: Stop emitting `PointsTo` from `MetaSlot::get_from_unused`

rikosellic:
I find that the `PointsTo` output of  `MetaSlot::get_from_unused` is not used anywhere, so this new design may be cleaner. @SNoAnd You may have to alter the definition in embedding accordingly, but that may be a low priority.


rikosellic:
The suspicious `sync_slot_perm` is also removed.

hiroki-chen (approved):
LGTM

PR #616, opened 2026-07-07
Prove fn page_size without external_body

Je5s1e:


rikosellic (changes requested):
This modification changes the source code. I think you can put all the definitions in a proof block and keep the original code at the end of the function.

rikosellic:

rikosellic (approved):

PR #617, opened 2026-07-07
Align dma_stream helper structure with mainline

Je5s1e:


hiroki-chen:
Please explain why is the snapshot for `DmaStreamInner` is justified:

- Why is better than wrapper method using `rwarc` locks?
- Who will provide the snapshot? Who tracks it? Who ensures that the snapshot is "the" current state of the dma stream?

To me this looks unsound.

hiroki-chen:
> Please explain why is the snapshot for `DmaStreamInner` is justified:
> 
> * Why is better than wrapper method using `rwarc` locks?
> * Who will provide the snapshot? Who tracks it? Who ensures that the snapshot is "the" current state of the dma stream?
> 
> To me this looks unsound.

On a sidenote this design is intentional because tracking the runtime properties via `RwArc` is not well-supported. Maintaining the invariant without reasoning about *how* invariant mutates provides little verification meanings

PR #622, opened 2026-07-12
fix: uninterp fn and axioms in `KernelPtConfig`

rikosellic:
Delete redundant fields and rename `FrameEntryOwner` to a ghost `FrameEntryState` as it does not carry any permissions.

### DIFF
--- a/ostd/specs/mm/page_table/node/entry_owners.rs
+++ b/ostd/specs/mm/page_table/node/entry_owners.rs
@@ -15,10 +15,8 @@ use crate::mm::page_prop::PageProperty;
 use crate::mm::page_table::*;
 use crate::mm::{Paddr, PagingConstsTrait, PagingLevel, Vaddr};
 use crate::specs::arch::*;
-use crate::specs::arch::{NR_ENTRIES, NR_LEVELS, PAGE_SIZE};
 use crate::specs::mm::frame::mapping::{frame_to_index, meta_addr};
-use crate::specs::mm::frame::meta_owners::PageUsage;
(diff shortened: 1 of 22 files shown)


ostd/specs/mm/page_table/node/entry_owners.rs
```diff
@@ -26,24 +26,18 @@ use core::marker::PhantomData;
...
-/// - `size` is the size of the mapping, which corresponds to the level of the parent
-///   page table. 4k at level 1, 2M at level 2, and 1G at level 3.
 /// - `prop` is a bitfield tracking the properties of the page ([`PageProperty`])
-/// - `is_tracked` refers to whether the frame is ref-counted (when `true`) or
-///   raw MMIO (when `false`).
-pub tracked struct FrameEntryOwner {
-    pub ghost mapped_pa: usize,
-    pub ghost size: usize,
-    pub ghost prop: PageProperty,
-    pub ghost is_tracked: bool,
```
hiroki-chen, line 41: Why is size unneeded here?

rikosellic: The only place `FrameEntryState` is used is in `EntryOwnerKind`, which is then used in `EntryOwner`. There is a `parent_level` field in that structure, and we can use `page_size(self.parent_level)` to get the page size.

rikosellic:
I find that currently we use `uninterp fn` in `impl PageTableConfig for KernelPtConfig`. This PR defines these specifications and removes several axioms. Sadly, we still have to retain `external_body` while proving the actual `item_from_raw`, because the design does not yet consider ' MetaRegionOwners '.

PR #626, opened 2026-07-13
fix `Cursor::new` that may create duplicate page table permissions

hiroki-chen:
This method does not take a page table owner which is unsound.

### DIFF
--- a/ostd/src/mm/page_table/cursor/locking.rs
+++ b/ostd/src/mm/page_table/cursor/locking.rs
@@ -40,6 +40,8 @@ pub assume_specification<Idx: Clone>[ Range::<Idx>::clone ](range: &Range<Idx>) 
         Tracked(regions): Tracked<&mut MetaRegionOwners>,
         Tracked(guards): Tracked<&mut Guards<'rcu>>
     requires
+        pt.relates_owner(pt_own, *old(regions)),
+        pt_own.0.value.node().relate_guard(root_guard),
         forall|i: int| 0 <= i < NR_ENTRIES ==> pt_own.0.children[i] is Some,
         va.start < va.end,
         // Per-config tightening; see `Cursor::new`. Pulled through to the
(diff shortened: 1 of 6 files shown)


ostd/src/mm/page_table/cursor/locking.rs
```diff
@@ -123,31 +130,65 @@ pub fn lock_range<'rcu, C: PageTableConfig, A: InAtomicMode>(
...
-    /*
-    let mut subtree_root = loop {
-        if let Some(subtree_root) = try_traverse_and_lock_subtree_root(pt, guard, va) {
-            break subtree_root;
+    // Retry if the candidate subtree became stray before we acquired its lock.
+    // A failed attempt is transactional at the specification level, so every
+    // iteration starts from the same cursor and ownership state.
+    let ghost cursor_own_before_retry = cursor_own;
+    let ghost regions_before_retry = *regions;
+    let ghost guards_before_retry = *guards;
```
rikosellic: Do we need these ghost variables?

ostd/src/mm/page_table/cursor/locking.rs
```diff
@@ -15,6 +18,7 @@ use crate::mm::{
 use vstd_extra::array_ptr::*;
 
 use crate::mm::page_table::*;
+use crate::specs::mm::frame::meta_owners::Metadata;
```
rikosellic: I prefer to aggregate this import with the next line.

ostd/src/mm/page_table/cursor/locking.rs
hiroki-chen: If removed, Verus cannot reason about that the code doesn't change anything in the loop invariant and I haven't found a better way to state this fact.

rikosellic: You may write `old(regions)` or `old(guards)` in the invariants. I'm also wondering whether we can use `#[verifier::loop_isolation(false)]` so we don't need this fact in the invariant.

hiroki-chen: Using `old` broke the verification for some weird reason in my previous attempts. I'll try if loop isolation works.

rikosellic:

PR #627, opened 2026-07-13
Prove mm tracked owner helper axioms

DID-Lab-SZU:


ostd/specs/mm/frame/linked_list/linked_list_owners.rs
```diff
@@ -164,8 +164,8 @@ impl Inv for LinkModel {
 }
 
 pub tracked struct LinkOwner {
```
rikosellic, line 166: Since `LinkOwner` does not contain any tracked permission. Can you try to rename it to a ghost struct `LinkState`, or even delete this struct?

rikosellic (changes requested):

ostd/specs/mm/frame/linked_list/linked_list_owners.rs
```diff
@@ -229,8 +229,8 @@ impl Inv for LinkedListModel {
 
 pub tracked struct LinkedListOwner<M: AnyFrameMeta + Repr<MetaSlotSmall>> {
```
rikosellic, line 230: And since `LinkOwner` is ghost, you can also make this one ghost, or deelte it.

ostd/specs/mm/frame/linked_list/linked_list_owners.rs
DID-Lab-SZU: If the `LinkOwner` must be ghost structs, this would require a larger refactor: CursorOwner and ListStore would also need to be ghostified, and the current tracked_insert / tracked_remove / tracked_take state updates would need to become ghost-state transitions. I think that is out of scope for this PR, which is focused on proving helper axioms.

rikosellic: It makes sense. I will leave it for future work.

rikosellic (approved):

rikosellic:

PR #632, opened 2026-07-15
Prove tracked MM ownership helper axioms

DID-Lab-SZU:


rikosellic:
Please merge the latest main.

ostd/specs/mm/embedding/mod.rs
```diff
@@ -538,9 +538,9 @@ pub enum CursorKind {
 /// cursor holds locked; mirrors what the exec `Cursor` carries via
 /// `path: [Option<PageTableGuard<'rcu, C>>; NR_LEVELS]`.
 pub tracked struct CursorEntry<'rcu> {
-    pub vm_space: VmSpaceId,
-    pub kind: CursorKind,
-    pub va: Range<Vaddr>,
+    pub ghost vm_space: VmSpaceId,
```
hiroki-chen, line 541: They should be renamed to `ghost_vm_space` etc.

rikosellic: Should we only add the `ghost_` and `tracked_` prefixes in the `exec struct`?

ostd/specs/mm/page_table/cursor/owners.rs
```diff
@@ -173,14 +185,18 @@ impl<'rcu, C: PageTableConfig> CursorContinuation<'rcu, C> {
...
     }
 
-    pub axiom fn tracked_new(
+    pub proof fn tracked_new(
         tracked owner_subtree: OwnerSubtree<C>,
         idx: usize,
-        guard: PageTableGuard<'rcu, C>,
+        tracked guard: PageTableGuard<'rcu, C>,
     ) -> (tracked res: Self)
         ensures
```
rikosellic: Please use `returns` here. And simplify the signature to `tracked Self`.

verified_libs/vstd_extra/src/ghost_tree.rs
```diff
@@ -305,6 +305,18 @@ impl<T: TreeNodeValue<L>, const N: usize, const L: usize> TreeNode<T, N, L> {
...
     }
 
+    /// Constructs a tracked node from its tracked fields.
+    pub proof fn tracked_new(
+        tracked value: T,
+        level: nat,
+        tracked children: Seq<Option<Self>>,
+    ) -> (tracked res: Self)
+        ensures
+            res == Self::new(value, level, children),
```
rikosellic: Use `returns` and simplify the signature.

ostd/specs/mm/embedding/mod.rs
hiroki-chen: Oh you are right!

rikosellic:

ostd/specs/mm/frame/meta_region_owners.rs
```diff
@@ -54,7 +54,7 @@ pub tracked struct MetaRegionOwners {
     /// [`crate::specs::mm::frame::segment::tracked_mint_seg_obligations`]).
     /// Multiset semantics — multiple outstanding obligations at the same slot
     /// are counted individually.
-    pub frame_obligations: vstd::multiset::Multiset<usize>,
+    pub ghost frame_obligations: vstd::multiset::Multiset<usize>,
```
rikosellic, line 57: @DID-Lab-SZU @hiroki-chen @Marsman1996. I find this modification is unsound. `frame_obligations` is used to track the `DropObligations` (which is also tracked) that are given to frames. If we change this field to ghost, then `MetaRegionOwners` will lose this track because it can be arbitrarily modified.

rikosellic: The correct way may be to use [GhostVarAuth](https://verus-lang.github.io/verus/verusdoc/vstd/resource/ghost_var/struct.GhostVarAuth.html), so we can ensure every consumption of the `DropObligation` is synchronized between the `Frame` and `MetaRegionOwner`.

rikosellic: Also see the comment: https://github.com/asterinas/vostd/blob/674043e7ba150bbd25e206e521ee80109dd13f5d/ostd/specs/mm/frame/meta_region_owners.rs#L274

PR #640, opened 2026-07-16
fix an unsound assumption while constructin PTEs

### LINKED ISSUE #624: `Child::into_pte` assumes the architecture-specific page-property  precondition
```rs
Child::Frame(paddr, level, prop) => {
    assume(C::E::new_page_req(paddr, level, prop));
    C::E::new_page(paddr, level, prop)
},
```

The assumption does not hold. Therefore the generic Child API claims any invariant frame child can become a valid PTE.

hiroki-chen:
Fix #624 as the current invariant is weak and allows us to construct invalid PTEs.

### DIFF
--- a/ostd/src/mm/page_table/mod.rs
+++ b/ostd/src/mm/page_table/mod.rs
@@ -214,6 +214,7 @@ pub unsafe trait PageTableConfig: Clone + Debug + Send + Sync + 'static {
             paddr % page_size(level) == 0,
             paddr + page_size(level) <= MAX_PADDR,
             Self::raw_item_well_formed(paddr, level, prop),
+            Self::E::new_page_req(paddr, level, prop),
         returns
             Self::item_into_raw_spec(item),
     ;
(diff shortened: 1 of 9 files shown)


ostd/src/mm/page_table/mod.rs
```diff
@@ -214,6 +214,7 @@ pub unsafe trait PageTableConfig: Clone + Debug + Send + Sync + 'static {
             paddr % page_size(level) == 0,
             paddr + page_size(level) <= MAX_PADDR,
             Self::raw_item_well_formed(paddr, level, prop),
+            Self::E::new_page_req(paddr, level, prop),
```
rikosellic, line 217: I'm not very sure about this. `new_page_req` was originally proposed only to restrict PTEs that map to pages. This condition requires all items to satisfy this constraint. Should we use `Self::Tracked(item) ==> Self::E::new_page_req(paddr, level, prop)`?

hiroki-chen: Tracked frames mean that they are refcounted but not all frames serve as normal pages. For example, MMIOs require us to create an untracked frame that is not subject to refcounting. If we are able to create invalid PTEs for MMIOs then the memory safety is not guaranteed.

rikosellic: I'm not familiar with the MMIO part. Does it mean that PTEs of untracked `MMIO` frames also behave as those referring to normal physical pages? So they should all follow the same `new_page_req` constraint?

hiroki-chen: Technically yes they should be the same backing physical pages except that the MMIO pages are free from ref counting.

rikosellic: Understood. Maybe we need to rename `new_page_req` to a more general name, like `new_leaf_pte_req`?

rikosellic: After a second thought, `new_page_req` will be okay.

rikosellic:

PR #641, opened 2026-07-16
Revert "Prove tracked MM ownership helper axioms (#632)"

rikosellic:
This reverts commit ab67553eb8579e3d6fd5baeca1c93c9483a030ff.


rikosellic:
I find #632 is unsound. See more discussions there.

PR #643, opened 2026-07-16
fix: proof error in (entry) owner spec

Marsman1996:
Verus, strange


rikosellic:
Thanks for your contributions！

PR #646, opened 2026-07-17
simplify: specs::mm::page_table::owners

Marsman1996:
2h10m

### DIFF
--- a/ostd/specs/mm/page_table/owners.rs
+++ b/ostd/specs/mm/page_table/owners.rs
@@ -114,12 +114,6 @@ pub proof fn lemma_vaddr_strict_bound(path: TreePath<NR_ENTRIES>)
     ensures
         vaddr(path) < 0x1_0000_0000_0000int,
 {
-    broadcast use {
-        TreePath::lemma_index_satisfies_elem_inv,
-        TreePath::lemma_push_tail_len,
-        TreePath::lemma_push_tail_preserves_inv,
-    };
-
     vstd::arithmetic::power2::lemma2_to64();
     vstd::arithmetic::power2::lemma2_to64_rest();
     if path.len() == 0 {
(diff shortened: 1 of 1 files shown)


ostd/specs/mm/page_table/owners.rs
```diff
@@ -120,52 +120,31 @@ pub proof fn lemma_vaddr_strict_bound(path: TreePath<NR_ENTRIES>)
...
     vstd::arithmetic::power2::lemma2_to64();
     vstd::arithmetic::power2::lemma2_to64_rest();
     if path.len() == 0 {
     } else if path.len() == 1 {
-        path.lemma_index_satisfies_elem_inv(0);
         let i0 = path[0];
-        assert(0 <= i0 < NR_ENTRIES);
         assert(rec_vaddr(path, 1) == 0);
     } else if path.len() == 2 {
-        path.lemma_index_satisfies_elem_inv(0);
```
rikosellic, line 133: So weird. I've tried to delete these `lemma_index_satisfies_elem_inv` calls myself, but then the verification failed. Now deleting more proofs makes it work again.

ostd/specs/mm/page_table/owners.rs
```diff
@@ -120,52 +120,31 @@ pub proof fn lemma_vaddr_strict_bound(path: TreePath<NR_ENTRIES>)
         TreePath::lemma_push_tail_preserves_inv,
     };
```
rikosellic: Do we need this broadcast use? It is already included in the `broadcast use group_ghost_tree_lemmas` at the beginning of the file.

rikosellic:
Please also merge the latest main.

ostd/specs/mm/page_table/owners.rs
Marsman1996: > I've tried

Maybe it because the update of Verus or the other lemmas

ostd/specs/mm/page_table/owners.rs
Marsman1996: No, we don't need, I will add this into the simplify script

rikosellic:

rikosellic (approved):

PR #647, opened 2026-07-17
chore: files cleanup and README update

Marsman1996:

PR #650, opened 2026-07-17
chore: delete more files

Marsman1996:
@rikosellic Actually I'm thinking about deleting the `osdk/` and `kernel/` since we might never prove these code


rikosellic:
> @rikosellic Actually I'm thinking about deleting the `osdk/` and `kernel/` since we might never prove these code

Actually, I thought you would delete them in #647.

Marsman1996:
> > @rikosellic Actually I'm thinking about deleting the `osdk/` and `kernel/` since we might never prove these code
> 
> Actually, I thought you would delete them in #647.

At that time I thought `ostd/` might be related to the `osdk_heap_allocator `, but actually there is no caller/callee relation...

rikosellic:

Marsman1996:
> Thanks for your contributions!

I haven't deleted `osdk/` and `kernel/` yet 👀

rikosellic:
Just make a new PR will be okay.

PR #658, opened 2026-07-20
fix: Redesign `TrackDrop`

rikosellic:
The current `ManuallyDrop<T: TrackDrop>` is unsound. It consumes the obligation immediately after `ManuallyDrop::new`, so the `Drop` can never be called because the obligation is leaked! This PR completely redesigns the `TrackDrop` infrastructure:
- `Key` and `DropObligation<Self::Key>` in `TrackDrop` are replaced by a general `Obligation` associated type, because `DropObligation` can be created from nowhere, violating the linear requirement of the permission if the user needs it.
- `ManuallyDrop<T: TrackedDrop>` now carries the obligation with it, instead of consuming it immediately, so that the obligation can be returned.
- `constructor_*/ consume_*` are removed because they are not used anymore.
- A `tracked_redeem(State) -> Obligation` method is added so that the state can emit an obligation.

Note1: The unsound `frame_obligation` design in `MetaRegionOwners` is not changed in this PR. Its replacement with fractional permission is left for a future PR.

Note2: This new design retains the ability to track whether a memory leak occurs. It is just the responsibility of the trait implementor, like before.

### DIFF
--- a/verified_libs/vstd_extra/src/drop_tracking.rs
+++ b/verified_libs/vstd_extra/src/drop_tracking.rs
@@ -16,17 +16,13 @@ verus! {
 ///
 /// 1. **Token (this type)**: an `ExclusiveGhost<R>` wrapper. Each `alloc`
 ///    produces a unique [`Loc`]; two outstanding tokens can be proven
-///    distinct via [`DropObligation::validate_with_other`]. No external_body
-///    axioms — the allocation is verified by `vstd`'s resource algebra.
+///    distinct via [`DropObligation::validate_with_other`].
 /// 2. **Ledger (in State)**: optional. For impls that opt in, the State
 ///    carries a set/multiset of outstanding keys; mint adds, redeem removes,
 ///    and the State's `clean_inv()` (or any boundary predicate) requires the
-///    set to be empty (or the per-slot counter to be zero). Combined with
+///    set to be empty. Combined with
 ///    `1`, the linear guarantee is sound: silently dropping the token leaves
 ///    the ledger non-empty and breaks the boundary check.
-///
-/// Impls that don't add a per-state ledger still pass tokens through —
-/// they get the API uniformity but not the enforcement.
 #[verifier::reject_recursive_types(R)]
 pub tracked struct DropObligation<R> {
     inner: ExclusiveGhost<R>,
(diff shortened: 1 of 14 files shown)


hiroki-chen:
The design looks good to me overall but there is one issue left regarding the independent executable-semantics issue. The `ManuallyDrop` still stores the `value: T` which means the vanilla Rust drop glue drops `T` when the wrapper goes out of the scope. The ghost obligation doesn't necessarily stop Rust field destruction. Should this be `core::mem::ManuallyDrop<T>` instead> Otherwise this is problematic.

verified_libs/vstd_extra/src/drop_tracking.rs
```diff
@@ -1,233 +1,175 @@
...
-/// At the time of this redesign no ostd callsite extracts a `Frame` from
-/// a `ManuallyDrop<Frame>` (only `Deref` borrows are taken; the one
-/// `into_inner` is on `MD<Arc<T>>`, not `MD<Frame>`). Adding such an
-/// extraction without the matching mint resurrects the double-counting
-/// bug that motivated this redesign.
-pub struct ManuallyDrop<T: TrackDrop>(pub T);
+} // verus!
+#[verus_verify]
+pub struct ManuallyDrop<T: TrackDrop> {
+    value: T,
```
hiroki-chen: Should it be `core::mem::ManuallyDrop<T>`?

rikosellic: Indeed! Although currently Verus's support for `Drop` is very poor, and we have to call all `drop`s manually, this is definitely a flaw that I failed to catch in the source code.

SNoAnd:
Looks good to me, removes a lot of redundancy! And having `ManuallyDrop` carry the obligation is correct.

> The design looks good to me overall but there is one issue left regarding the independent executable-semantics issue. The `ManuallyDrop` still stores the `value: T` which means the vanilla Rust drop glue drops `T` when the wrapper goes out of the scope. The ghost obligation doesn't necessarily stop Rust field destruction. Should this be `core::mem::ManuallyDrop<T>` instead> Otherwise this is problematic.

Do we plan to use an executable semantics on the Verus code? If so this makes a lot of sense, and I will follow similar patterns in the future to make sure that the behavior matches.

rikosellic:
I think we should keep the behaviour aligned with Rust. And making this change makes it eaiser to migrate if Verus adds more support for `Drop` and `ManuallyDrop` later.

PR #663, opened 2026-07-21
fix: Use fractional permissions in `Frame`, remove `frame_oblilgations`

rikosellic:
I find that instead of trying to make `DropObligations` sound based on some ghost resource tokens, it is easier to directly use fractional permissions. This PR proposes the following change:
- Record the fractional permission `Count<simple_pptr::PointsTo<MetaSlot>, REF_COUNT_MAX>` in the `Frame` definition, so it is valid to access the fields directly without accessing `MetaRegionOwners`. (Accessing the actual value still requires `MetaRegionOwners`).
- Remove the unsound `frame_obligations` as the counting is already expressed by the fractional permission. Change the value of `slots` to `CountResource` accordingly.
- Remove `impl TrackDrop for Frame` and our customized `ManuallyDrop<Frame>` because the permission is already stored in `Frame`. We may still need a way to take out the permission, though. So it may be wrapped with an `Option`.

This will be a big change, but I think this is a step forward to concurrent verification.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -589,11 +589,11 @@ dependencies = [
 
 [[package]]
 name = "verus_builtin"
-version = "0.0.0-2026-05-17-0151"
+version = "0.0.0-2026-07-27-0206"
 
 [[package]]
 name = "verus_builtin_macros"
-version = "0.0.0-2026-07-12-0122"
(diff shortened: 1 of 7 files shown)


rikosellic:
This change is too complex without a unified `MetaRegionOwners`. I will try to do that first.

PR #668, opened 2026-07-22
refactor: Remove `Metadata`

rikosellic:
Currently, there is an exec `MetaData` struct that does not exist in Asterinas source code, serving as a bridge between `MetaSlot` and `M: AnyFrameMeta` conversion. This PR removes this indirection and establishes direct translation between `MetaSlotStorage` and `M: AnyFrameMeta`. I believe this makes the model a bit cleaner and makes the phase III memory model easier to integrate.

Main changes include:
- [meta_owners.rs] Delete MetaData. 
- [linked_list_owners.rs/ linked_list.rs] Replace `ReprPtr<MetaSlot, MetadataAsLink<M>` with  `ReprPtr<MetaSlotStorage, Link<M>`
- The `ReprPerm` that enables the type casting is added to the `UniqueFrameOwners`, etc, not in `MetaRegionOwners`.

Note: The metadata operations are now proved or axiomatized in `meta_owners.rs`, including `borrow_meta`, `borrow_meta_mut` and `write_metadata_in_storage`. `Frame::as_meta_ptr` is disabled for now. They should be further investigated, modeled, and proved in phase III.

### DIFF
--- a/ostd/specs/mm/frame/meta_owners.rs
+++ b/ostd/specs/mm/frame/meta_owners.rs
@@ -200,49 +200,45 @@ pub tracked struct MetadataInnerPerms {
     pub in_list: PermissionU64,
 }
 
-/// Reconstructs the representation permission paired with a concrete view of
-/// the type-erased metadata storage.
-pub uninterp spec fn repr_perm_from_storage<M: AnyFrameMeta + Repr<MetaSlotStorage>>(
-    perm: pcell_maybe_uninit::PointsTo<MetaSlotStorage>,
-) -> M::Perm;
+/// Well-formedness of a concrete metadata representation. The permissions for
+/// the outer slot and its storage remain owned by `MetaRegionOwners`; only the
+/// representation-specific permission is supplied by the concrete metadata
+/// owner.
+pub open spec fn typed_meta_wf<M: AnyFrameMeta + Repr<MetaSlotStorage>>(
+    points_to: vstd::simple_pptr::PointsTo<MetaSlot>,
+    storage: pcell_maybe_uninit::PointsTo<MetaSlotStorage>,
+    repr_perm: M::Perm,
+) -> bool {
+    &&& points_to.is_init()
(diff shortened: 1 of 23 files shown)


rikosellic:
@hiroki-chen @SNoAnd please take a look!

ostd/specs/mm/frame/meta_owners.rs
```diff
@@ -193,6 +200,84 @@ pub tracked struct MetadataInnerPerms {
...
+        final(slot_owner).paths_in_pt == old(slot_owner).paths_in_pt,
+        final(slot_owner).inner_perms.ref_count == old(slot_owner).inner_perms.ref_count,
+        final(slot_owner).inner_perms.vtable_ptr == old(slot_owner).inner_perms.vtable_ptr,
+        final(slot_owner).inner_perms.in_list == old(slot_owner).inner_perms.in_list,
+        typed_meta_wf::<M>(*points_to, final(slot_owner).inner_perms.storage, *final(repr_perm)),
+        *final(res) == typed_meta_value::<M>(
+            final(slot_owner).inner_perms.storage,
+            *final(repr_perm),
+        ),
+{
```
rikosellic, line 271: This function looks pretty unsound, but it is as unsound as the original `borrow_mut_typed_perm`. Maybe it can only be resolved when we have full `dyn` support.

rikosellic: Turns out to be a Verus bug, and it has been fixed. Then the only unsound part is `Repr::from_borrowed_mut`. But that unsoundness already exists in the old design and can only be resolved with `dyn`. BTW, this PR has already removed a lot of `external_body`s and axioms.

hiroki-chen (commented):
LGTM!

SNoAnd:
This looks good to me! I've been working through the foundations of this in VerusBelt, and I do think these changes will simplify the conversion later.

SNoAnd (approved):

PR #672, opened 2026-07-27
Cpu local model

hiroki-chen:
`CpuCore` is a spec-only model for the current CPU core and the handle is the linear resource you'd "open" the current core for verification so you can store anything inside a CPU core.

### DIFF
--- a/ostd/specs/task/cpu_core.rs
+++ b/ostd/specs/task/cpu_core.rs
@@ -1,23 +1,23 @@
 // SPDX-License-Identifier: MPL-2.0
-//! Proof model for one CPU core.
+//! Proof model for ownership of one CPU's local resources.
 //!
-//! A [`CpuCore`] permanently owns the CPU-local resources assigned to one
-//! logical CPU. Scheduling changes only the core's `current_task`; it never
+//! A [`CpuCoreOwner`] permanently owns the CPU-local resources assigned to one
+//! logical CPU. Scheduling changes only the owner's `current_task`; it never
 //! transfers those resources to the task. Runtime CPU-local access temporarily
-//! opens the core into a linear [`CpuCoreHandle`] and its typed local state,
-//! then restores that state before returning the core to the scheduler.
+//! opens the owner into a linear [`CpuCoreOwnerHandle`] and its typed local
+//! state, then restores that state before returning the owner to the scheduler.
 use core::marker::PhantomData;
 
 use vstd::{prelude::*, resource::Loc};
 use vstd_extra::resource::ghost_resource::excl::ExclusiveGhost;
 
 use crate::specs::mm::cpu::CpuId;
-use crate::specs::task::cpu_local::CpuLocalInstance;
+use crate::specs::task::cpu_local::CpuLocalAuth;
 
 verus! {
 
-/// Logical state of a CPU core that is independent of its CPU-local payload.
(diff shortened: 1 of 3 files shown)


rikosellic:
The definitions are consistent and well-defined, but I'm unclear on how they can be used for verification.

BTW, what's the advantage of using `Seq<Loc>` over `Set<Loc>` for `local_keys`?

hiroki-chen:
`Seq<Loc>` was used to preserve order and multiplicity, so that two same-typed resources cannot be exchanged during open/restore. A `Set<Loc>` would identify (a, b) and (b, a). However, ordering is probably not semantically relevant for CPU-local objects. This suggests that `local_key` is compensating for an overly general `CpuCore<L>` interface.

hiroki-chen:
The intended verification workflow is for the scheduler to own `CpuCore<L>`. A preemption-disabled CPU-local access temporarily opens it into `CpuCoreHandle<L>` and `L`, uses theelevant points-to resource, and then restores the same typed aggregate with the same ordered identities. A schedule-out requires the complete core, so it cannot occur while the CPU-local state is open.

rikosellic:
Yes, it is indeed very general and confuses me a bit. Is `CpuLocalInstance` owned by the scheduler? And there will be more than one of these, otherwise, the `id` as `local_key` does not make sense.

hiroki-chen:
`CpuLocalInstance` is not intended to be scheduler-owned but an authoritative state for one logicasl CPU-local object and should be stored in that object's invariant. The scheduler/CPU-core side owns the corresponding per-CPU permissions.

There is one authoritative instance per CPU-local object, not one per CPU. For example, `FOO` and `BAR` can have different authority IDs, while `FOO`'s points-to resources on all CPUs share its authority ID. A CPU core containing both resources therefore has a key like `[foo_id, bar_id]`. Accessing FOO on CPU 0 combines the authority from FOO's invariant with the FOO@CPU0 points-to resource from CPU 0's core state. The name `CpuLocalInstance` could be a little bit misleading here; `CpuLocalAuth` would better describe its role.

rikosellic:
Thanks! Your explanation makes the design very clear. My only question now is whether we should use the `CpuCoreLocalState` to aggregate the permissions with `locals_key` in each `CpuCore`. Can we just define a struct that records all the resources?

hiroki-chen:
This `CpuCoreLocalState` is intended to be an interface implemented by a concrete tracked struct that needs to store all per-core resources. For example, an `OstdCpuLocals` struct can contain the RCU, timer, and preemption points-to resources and a scheduler can own `CpuCore<OstdCpuLocals>`.

`local_key` should therefore be derived from those fields that rather than stored redundantly in the aggregate. Its purpose is only to let the handle verify that the same resources finally return to their original locations. If OSTD ultimately has exactly one fixed CPU-local aggregate (I'm not very sure about it so I reserve an interface for future usages), we could remove the trait and make `CpuCore` directly own that concrete struct.

rikosellic:
Indeed, we will never know whether this design is appropriate until we start to verify things. Please change the names of `CpuCore` and `CpuLocalInstance` though. `CpuCore` is too general.I will merge it then.

rikosellic:
Maybe `CpuCoreTrackedState` or `CpuCoreOwner`?

rikosellic:

PR #673, opened 2026-07-27
fix: Use `&'static PointsTo<MetaSlot>` in `MetaRegionOwners.slots`

rikosellic:
Replace `MetaRegionOwners.slots` with  `Map<int, &'static PointsTo<MetaSlot>>` and align `MetaSlot::get_slot` with the Asterinas source code. It is much easier than I thought.

### DIFF
--- a/ostd/src/mm/frame/linked_list.rs
+++ b/ostd/src/mm/frame/linked_list.rs
@@ -20,7 +20,7 @@ use crate::mm::kspace::FRAME_METADATA_RANGE;
 use crate::specs::arch::*;
 use crate::specs::mm::frame::{
     linked_list::linked_list_owners::*,
-    mapping::{frame_to_index, group_page_meta, index_to_meta, max_meta_slots, meta_to_index},
+    mapping::{frame_to_index, group_page_meta, index_to_meta, meta_to_index},
     meta_owners::{
         MetaSlotOwner, MetaSlotStorage, borrow_meta, borrow_meta_mut, typed_meta_value,
         typed_meta_wf,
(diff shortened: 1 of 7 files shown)


ostd/src/mm/frame/linked_list.rs
```diff
@@ -31,7 +31,7 @@ use crate::specs::mm::frame::{
 
 use super::{
     MetaSlot, mapping,
-    meta::{AnyFrameMeta, get_slot},
+    meta::{_VERUS_VERIFIED_get_slot, AnyFrameMeta, get_slot},
```
hiroki-chen: Don't import this. Instead, do a qualified call like `#[verus_spec] a::b::c::d()`. I think should be hardcoded to LLM's behavior.

hiroki-chen (changes requested):

hiroki-chen:
Other things do LGTM.

ostd/src/mm/frame/linked_list.rs
rikosellic: It is already called with `#[verus_spec]`, I'm not sure why this happens.
```
let Ok(slot) = (#[verus_spec(with Tracked(slot_perm))]
        get_slot(frame))
```

rikosellic: Anyway, using `crate::mm::frame::meta::get_slot`works. Thx!

hiroki-chen: > Anyway, using `crate::mm::frame::meta::get_slot`works. Thx!

This is a known limitation of `verus_spec` because you'd mention if by `path` not by its name:

```rs
#[verus_spec(with xxx)] a(); // fails
#[verus_spec(with xxx)] some_mod::a(); // works

**Review by hiroki-chen — APPROVED** [review]

File: ostd/src/mm/frame/linked_list.rs [hunk 2]
```diff
@@ -394,37 +394,27 @@ impl<M: AnyFrameMeta + Repr<MetaSlotSmall>> LinkedList<M> {
             old(owner).list_id != 0 ==> *final(owner) == *old(owner),
     )]
     pub fn contains(&mut self, frame: Paddr) -> bool {
-        let Ok(slot_ptr) = get_slot(frame) else {
+        proof_decl! {
+            let ghost idx = frame_to_index(frame);
+                if valid_frame_paddr(frame) {
```
SNoAnd: Weird indentation here.

SNoAnd (approved):
Looks good to me, other than the tiny formatting issue.

PR #674, opened 2026-07-28
 Prove page table stores and tracked owner helpers

DID-Lab-SZU:
Prove some functions marked with #[verifier::external_body]


rikosellic:
Please merge the latest main, thanks!

ostd/src/mm/kspace/kvirt_area.rs
```diff
@@ -56,14 +56,12 @@ pub open spec fn frame_as_dynframe<T: AnyFrameMeta + Repr<MetaSlotStorage>>(
...
-/// to the spec function `frame_as_dynframe`. The `Into` impl uses `transmute`, so the
-/// function is marked `external_body` — same trust boundary as the underlying conversion.
-#[verifier::external_body]
+/// to the spec function `frame_as_dynframe`.
 fn frame_into_dynframe<T: AnyUFrameMeta>(frame: Frame<T>) -> (res: DynFrame)
     ensures
         res == frame_as_dynframe(frame),
 {
-    frame.into()
+    DynFrame { ptr: frame.ptr, _marker: PhantomData }
```
Marsman1996, line 65: Question: what is the original exec code here?

DID-Lab-SZU: I changed frame_into_dynframe to construct DynFrame directly so the conversion is explicit and easier for Verus to verify. The old frame.into() call relied on trait resolution and an implicit zero-cost conversion, which made the proof less transparent. This does not change semantics: DynFrame is just a repr(transparent) wrapper around the same ptr and zero-sized marker fields, so the new form preserves the exact runtime representation and behavior.

Marsman1996: keep the original exec code as the comment maybe

```suggestion
    /* frame.into() */
    DynFrame { ptr: frame.ptr, _marker: PhantomData }
```

verified_libs/vstd_extra/src/array_ptr.rs
```diff
@@ -529,6 +529,20 @@ pub exec fn layout_for_array_is_valid<V: Sized, const N: usize>()
...
 
 impl<V, const N: usize> ArrayPtr<V, N> {
+    /// Reconstructs a pointer to the selected array element.
+    #[inline(always)]
+    pub exec fn as_mut_ptr(&self, Tracked(perm): Tracked<&PointsTo<V, N>>) -> (res: *mut V)
+        requires
+            perm.wf(),
+            perm.is_pptr(*self),
+            self.index < N,
+    {
```
rikosellic, line 543: Why does this function not have any postcondition?

ostd/src/mm/page_table/mod.rs
```diff
@@ -1830,26 +1830,34 @@ pub unsafe fn load_pte<E: PageTableEntryTrait>(
     unimplemented!()
 }
 
+/// Verus specification for reconstructing an atomic reference from an aligned `usize` pointer.
+#[doc(hidden)]
+pub assume_specification<'a>[ AtomicUsize::from_ptr ](ptr: *mut usize) -> &'a AtomicUsize
+;
+
```
rikosellic: Move this into `vstd_extra/external`

ostd/src/mm/page_table/mod.rs
```diff
@@ -1859,6 +1867,14 @@ pub unsafe fn store_pte<E: PageTableEntryTrait>(
...
     ordering: Ordering,
-);
+) {
+    proof {
+        ptr.tracked_overwrite(perm, ptr.index, new_val);
+    }
+    let ptr = ptr.as_mut_ptr(Tracked(&*perm));
+    let new_raw = new_val.as_usize();
+    let atomic = unsafe { AtomicUsize::from_ptr(ptr.cast()) };
+    atomic.store(new_raw, ordering);
```
rikosellic: Is this `store` function defined in `vstd`?

DID-Lab-SZU: `store` itself is defined by `core::sync::atomic::AtomicUsize`, not by vstd. vstd defines a verified atomic wrapper with a `store` method, but this code uses the standard-library atomic reference reconstructed by `AtomicUsize::from_ptr`. I moved the `from_ptr` assumption into `vstd_extra::external::atomic` so the std atomic boundary is centralized there.

rikosellic: Then it is weirder. Is this `store` method even verified? I'm surprised it passes verification.

Marsman1996 (approved):
Since you revert that commit, I think this is Ok to me

verified_libs/vstd_extra/src/external/atomic.rs
```diff
@@ -0,0 +1,10 @@
+use core::sync::atomic::AtomicUsize;
+use vstd::prelude::*;
+
+verus! {
+
+/// Specification for `AtomicUsize::from_ptr`.
+pub assume_specification<'a>[ AtomicUsize::from_ptr ](ptr: *mut usize) -> &'a AtomicUsize
```
rikosellic: Is this function still used? I believe we'd better remove it if it is no longer used, because it may be unsound to add this unsafe function without any pre-conditions.

rikosellic:
We are actively working on a version update. So the merge of this PR will be poseponed until #678 is merged to prevent any unexpected failures. The modification LGTM.

rikosellic:
Please merge the latest main.

rikosellic:
MacOS fails again..

Marsman1996:
So this time succeed in macOS and fail in Ubuntu...

Marsman1996 (approved):
Since all CIs have been passed, LGTM
@rikosellic

rikosellic:

PR #677, opened 2026-07-30
Prove tlb flush functions

DID-Lab-SZU:


ostd/src/mm/tlb.rs
```diff
@@ -322,7 +347,13 @@ impl OpsStack {
             self.size = 0;
             return;
         }
-        for i in 0..other.size {
+        let old_size = self.size;
+        for i in 0..other.size
+            invariant
+                self.size == old_size + i,
```
Marsman1996: Can we use old(self.size) here?

ostd/src/mm/tlb.rs
```diff
@@ -217,11 +238,13 @@ impl TlbFlushOp {
...
 
-    #[verifier::external_body]
     fn optimize_for_large_range(self) -> Self {
         match self {
             TlbFlushOp::Range(range) => {
-                if range.len() > FLUSH_ALL_RANGE_THRESHOLD {
+                // Equivalent to `range.len() > FLUSH_ALL_RANGE_THRESHOLD`:
+                // `Range::<usize>::len` is `end.checked_sub(start).unwrap_or(0)`,
+                // i.e. saturating subtraction (never panics on inverted ranges).
+                if range.end.saturating_sub(range.start) > FLUSH_ALL_RANGE_THRESHOLD {
```
Marsman1996: `range.end.saturating_sub(range.start)` is the `range.len()`, why we have to rewrite here

ostd/src/mm/tlb.rs
```diff
@@ -185,9 +204,11 @@ impl<'a /*, G: PinCurrentCpu*/> TlbFlusher<'a /*, G*/> {
 
         self.have_unsynced_flush = CpuSet::new_empty();
          */
+
     }
 }
 
+} // verus!
```
Marsman1996: We can just delete this line and next `verus! {` to make them a whole `verus! {}` block

ostd/src/mm/tlb.rs
```diff
@@ -40,16 +40,18 @@ pub struct TlbFlusher<'a  /*, G: PinCurrentCpu*/ > {
 }
 
 } // verus!
+verus! {
```
Marsman1996: Also delete this line and the above line

ostd/src/mm/tlb.rs
```diff
@@ -65,13 +67,21 @@ impl<'a /*, G: PinCurrentCpu*/> TlbFlusher<'a /*, G*/> {
...
             *final(model) == old(model).issue_tlb_flush(op),
             final(model).inv(),
     )]
-    #[verifier::external_body]
     pub fn issue_tlb_flush(&mut self, op: TlbFlushOp) {
-        self.ops_stack.push(op, None);
+        // Update the ghost TLB model to reflect the pending flush (proof), then
+        // record the operation on the executable ops stack. The proof consumes
+        // `op` (matching the `ensures`), so `push` records a clone.
+        let op_clone = op.clone();
```
Marsman1996: Do we have to add a clone here?

Marsman1996 (changes requested):
I'm not sure, too many exec code modifications

Marsman1996 (approved):
From [fa2a3b8](https://github.com/asterinas/vostd/pull/677/commits/fa2a3b863da855c05dc47c690bb2b7f47c077d35) we can see that KVerus now still let the exec code modification passes without human review, which we shall think of a better method to guard the exec code...

hiroki-chen:
This should be closed because:

1. The TLB is CPU-local and should be explicitly modeled after CPU core model has been stablized.
2. The TLB module has no spec at all so the verification itself is at most vacuous.

@rikosellic I don't think this PR should be merged right now. Perhaps this is part of the phase III content?

hiroki-chen:
This will also interfere with page table designs because old mappings will likely need to reconcile with the latest modifications, for example.

rikosellic:
Yeah we should discuss this later.

PR #678, opened 2026-07-31
fix: adapt proofs for latest Z3 4.16.0

rikosellic:
Several proofs fail because Verus now uses the latest Z3. This PR stabilizes the proofs and tries to keep the verification time. `asterinas/verus` has not been updated yet.


rikosellic:
I haven't pushed five modifications about `rlimit` yet. The upstream test will fail because of rlimits. Adding `#[verifier::rlimit(400)]` will solve the issues, but that is not an ideal solution.

PR #679, opened 2026-07-31
prove: replace rlimit with proof helper function

Marsman1996:


ostd/src/mm/frame/linked_list.rs
```diff
@@ -149,7 +150,43 @@ proof fn meta_region_inv_at(regions: MetaRegionOwners, i: int)
...
         regions.slot_owners[i].slot_vaddr == regions.slots[i].addr(),
 {
-    reveal(<MetaRegionOwners as Inv>::inv);
+}
+
+/// Localizes the "no existing slot aliases the inserted frame" universal fact
+/// into its own prover query, so the `MetaRegionOwners::inv` and map-insert
+/// quantifiers do not get over-instantiated inside `insert_before`'s body.
+#[verifier::spinoff_prover]
+proof fn insert_before_slot_distinct<M: AnyFrameMeta + Repr<MetaSlotSmall>>(
```
rikosellic: Should we add a lemma prefix? There seem to be several functions missing the prefix in this file.

rikosellic (changes requested):

ostd/src/mm/frame/linked_list.rs
hiroki-chen: I remember someone said that if a function is not exposed as public then prefix is not required. BTW there is a coding style.

Marsman1996: How about the functions in `ostd/specs/mm/embedding/mod.rs`?
For example, the `pub proof fn insert_vm_space`, shall I add `lemma_` prefix for all of them?

rikosellic: I think we should all `lemma` to all these functions to make a consistent style.

Marsman1996:
Amazing Verus...

Marsman1996:
The proof code is considered redundant in this version of Verus, but it may become necessary again in future versions.

rikosellic:

rikosellic (approved):

PR #681, opened 2026-08-01
fix: error caused by Verus update

Marsman1996:
Fix the verification error caused by https://github.com/verus-lang/verus/pull/2736


rikosellic:

Marsman1996:
I brought `#[verifier::rlimit(200)]` back since the rlimit error has occurred again.

Maybe we should wait until Verus becomes more stable before applying the rlimit and proof code stripping.

rikosellic:
The rlimit change looks reasonable to me. The Z3 upgrade is a very rare situation, and I do not anticipate that Verus will be significantly more stable in the near future (years, maybe). We'd better continue proof code stripping.

PR #686, opened 2026-08-03
refactor: add `MetadataPerms` perms and remove `MetadataInnerPerms`

rikosellic:
This PR makes some preparations for #675. 

`MetadataInnerPerms` currently stores all permissions, but these four fields actually have two kinds of lifetimes:
- `MetaSlotStorage` and `vtable_ptr` can be allocated and reused according to the reference count.
- `in_list` and `ref_count` always live when the kernel is alive (i.e., 'static).

This PR moves `PointsTo<MetaSlotStorage>` and `vtable_ptr` into a separate `MetadataPerms` struct:
```
pub tracked struct MetadataPerms {
    pub storage: pcell_maybe_uninit::PointsTo<MetaSlotStorage>,
    pub vtable_ptr: simple_pptr::PointsTo<usize>,
}
```
`in_list` and `ref_count` are moved into `MetaSlotOwner`. `MetadataInnerPerms` is removed.

This PR only move things around and does not modify any actual permissions.

### DIFF
--- a/ostd/specs/mm/embedding/cursor.rs
+++ b/ostd/specs/mm/embedding/cursor.rs
@@ -109,9 +109,7 @@ pub axiom fn vm_space_cursor_embedded<'a, 'rcu>(
         final(regions).slots == old(regions).slots,
         forall|i: int|
             #![trigger final(regions).slot_owners[i]]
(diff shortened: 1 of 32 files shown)


Marsman1996:
I'm working for a method without adding `rlimit(600)`

PR #699, opened 2026-08-08
prove: util::range_alloc

Marsman1996:
~The 3 `#[verifier::external_body]` are caused by the `BTreeMap`~
This PR is parsed into 3 (including this one):
- [x] https://github.com/asterinas/vostd/pull/703 with `#[verifier::external_body]`
- [x] https://github.com/asterinas/vostd/pull/704 to add btree in vstd_extra


rikosellic:
I think vstd already supports `BTreeMap`? https://github.com/verus-lang/verus/pull/2189/files

Marsman1996:
> I think vstd already supports `BTreeMap`? https://github.com/verus-lang/verus/pull/2189/files

~But there is no spec foe `BTreeMap::get_mut`~
Well, I shall add them in the `vstd_extra`

rikosellic:
You may also create a PR for Verus.

rikosellic:
~~I will try to investigate `BTreeMap` more closely tomorrow. Then we can propose a PR for Verus. This PR should not be merged until the Verus PR is merged. It is not very urgent.~~

ostd/specs/util/range_alloc_specs.rs
rikosellic, line 1: Do we need this file, since the model is very small? Will this be extended if we verify more code about `range_alloc`?

Marsman1996:
716K used · 35M in · 88.7K out

ostd/specs/util/range_alloc_specs.rs
Marsman1996: No, it will not be extended. Actually, there is no caller for them in OSTD at all.

It is just because skill told agent to "move specs into `spec/` as much as possible", maybe we shall tell it to do it unless the model code is large (e.g., > 200 LoC)?

rikosellic: > maybe we shall tell it to do it unless the model code is large (e.g., > 200 LoC)?

Sounds good to me

verified_libs/vstd_extra/src/external/btree.rs
```diff
@@ -0,0 +1,148 @@
...
+            new_map,
+            key,
+            old_value,
+            new_value,
+        ) <==> {
+            &&& old_map.contains_key(*key)
+            &&& old_map[*key] == old_value
+            &&& new_map == old_map.insert(*key, new_value)
+        },
+;
```
rikosellic: Why do we need this function and axiom? Does this just mean `insert`?

verified_libs/vstd_extra/src/external/btree.rs
```diff
@@ -0,0 +1,148 @@
...
+>[ BTreeMap::<Key, Value, A>::get_mut::<Q> ](
+    map: &'a mut BTreeMap<Key, Value, A>,
+    key: &Q,
+) -> (result: Option<&'a mut Value>)
+    ensures
+        obeys_cmp::<Key>() ==> match result {
+            Some(value) => borrowed_key_mutated(old(map)@, final(map)@, key, *value, *final(value)),
+            None => !contains_borrowed_key(old(map)@, key) && final(map)@ == old(map)@,
+        },
+;
```
rikosellic, line 221: It is incorrect that this function does not have any preconditions. According to [the std documentation](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.get_mut), the ordering of `Q` must match the ordering of `Key`. This precondition is not expressed here.

verified_libs/vstd_extra/src/external/btree.rs
Marsman1996: This axiom provides a concrete interpretation of `borrowed_key_mutated` when the borrowed lookup key has the same type as the key stored in the map.

  The generic `BTreeMap::get_mut` specification must support borrowed-key lookups where the types differ, such as `BTreeMap<String, V>::get_mut(&str)`. In that case, the borrowed key cannot generally be converted back into the exact stored key, so the specification uses the abstract relation `borrowed_key_mutated`.

  When the borrowed and stored key types are both Key, the relation can be characterized precisely:

  - The key existed in `old_map`.
  - Its value was `old_value`.
  - `new_map` is `old_map` with that key mapped to `new_value`.

  Therefore, this is not specifying a call to executable `BTreeMap::insert`. The spec-level `Map::insert` is a functional update used to describe the final mathematical state of the map. The axiom also records the successful lookup and the previous value.

  Declaring it as a broadcast axiom allows Verus to apply this interpretation automatically when reasoning about `get_mut` with identical borrowed and stored key types.

rikosellic: This definition is very weird. Because the whole system only works when `Q` is identical to `Key`. A more general way may be to add a `borrow_spec` to the `Borrow` trait. Then we can have some definition like `exists |k: Key|, old_map.contains(key) && key.borrow_spec() == q && new_map == old_map.insert(k, new_value)`. Next we can add a broadcast lemma that handles the identical case, which removes the exists quantifier. However, this will require a fundamental change to `Borrow`'s spec, which is undefined now. You can find it in [`std_specs/core.rs`](https://github.com/verus-lang/verus/blob/92f466f247f45128c630d1c843fd6e27d2115587/source/vstd/std_specs/core.rs#L118-L121). You'd better open a question about how to support the general Borrow trait in Verus Zulip first.

verified_libs/vstd_extra/src/external/btree.rs
rikosellic: And I find this one is also related to `borrow_spec`. Maybe we can merge the version without `btree_map::CursorMut` and `BTreeMap::get_mut` first. The `borrow_spec` needs to be discussed with Verus developers

Marsman1996: > Maybe we can merge the version without `btree_map::CursorMut` and `BTreeMap::get_mut` first.

Yes, I agree.

Marsman1996:
Verus's external specification of Borrow<Q> currently has no logical model for this compatibility, so have to add a axiom in verified_libs/vstd_extra/src/external/btree.rs

verified_libs/vstd_extra/src/external/btree.rs
```diff
@@ -0,0 +1,168 @@
...
+pub uninterp spec fn borrowed_key_ordering_matches<Key, Q: ?Sized>(key: &Q) -> bool;
+
+/// A key type has the same ordering as itself.
+pub broadcast axiom fn axiom_deref_key_ordering_matches<Key>(key: &Key)
+    ensures
+        #[trigger] borrowed_key_ordering_matches::<Key, Key>(key),
+;
+
+/// Whether a stored key matches a borrowed lookup key.
+pub open spec fn borrowed_key_matches<Key, Q: ?Sized>(stored_key: Key, key: &Q) -> bool {
```
rikosellic: This function is redundant.

verified_libs/vstd_extra/src/external/btree.rs
```diff
@@ -0,0 +1,168 @@
...
+) -> bool {
+    &&& contains_borrowed_key(old_map, key)
+    &&& contains_borrowed_key(new_map, key)
+    &&& maps_borrowed_key_to_value(old_map, key, old_value)
+    &&& maps_borrowed_key_to_value(new_map, key, new_value)
+    &&& exists|remainder: Map<Key, Value>|
+        {
+            &&& borrowed_key_removed(old_map, remainder, key)
+            &&& borrowed_key_removed(new_map, remainder, key)
+        }
```
rikosellic, line 153: Why do we need this exists clause?

Marsman1996: The existential clause is the frame condition for `get_mut`. The preceding predicates only constrain the value associated with the borrowed key; without this clause, the specification would also allow arbitrary changes to every other entry. Requiring both maps to produce the same remainder after removing the borrowed key ensures that they differ only at that entry. We use `borrowed_key_removed` because, for a general Key: `Borrow<Q>`, Verus currently has no borrow_spec that lets us directly identify the corresponding stored Key.

For the `Q = Key` case used by `range_alloc`, the existential clause reduces to:

```rust
old_map.remove(*key) == new_map.remove(*key)
```

Together with the two value-mapping predicates, this states that only the value corresponding to `key` may change; all other entries remain unchanged. This matches [`BTreeMap::get_mut`](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.get_mut), which returns a mutable reference to the value corresponding to the borrowed key.

The `borrowed_key_ordering_matches` precondition separately captures the documented requirement that the ordering of `Q` must match the ordering of `Key`.

For an arbitrary heterogeneous `Key: Borrow<Q>`, this remains an abstract specification because vstd does not yet formalize the semantics and ordering laws of a general `Borrow` implementation. However, it is precise for the `Q = Key` case used here.

ostd/src/util/range_alloc.rs
```diff
@@ -20,6 +22,20 @@ pub struct RangeAllocError;
 
 verus! {
 
+broadcast use {group_btree_extra_axioms, vstd::std_specs::btree::group_btree_axioms};
+
+/// Spec model capturing the managed full range of a [`RangeAllocator`].
+pub ghost struct RangeAllocatorModel {
```
rikosellic: Is this still used?

ostd/src/util/range_alloc.rs
```diff
@@ -1,5 +1,8 @@
 // SPDX-License-Identifier: MPL-2.0
 use vstd::prelude::*;
+use vstd_extra::debug_assert;
+use vstd_extra::external::btree::*;
+use vstd_extra::panic::{UnwrapOrPanic, may_panic};
```
rikosellic: I prefer to aggregate this in one `vstd_extra::{...}`.

ostd/src/util/range_alloc.rs
```diff
@@ -54,9 +59,11 @@ impl RangeAllocator {
     }
 
     /// Allocates a specific kernel virtual area.
-    #[verifier::external_body]
     #[verus_spec(res =>
-        requires allocate_range.start < allocate_range.end,
+        requires
+            allocate_range.start < allocate_range.end,
+            self@.start <= allocate_range.start,
+            allocate_range.end <= self@.end,
```
rikosellic: Better to use a chaining operator to state this in one line.

ostd/src/util/range_alloc.rs
```diff
@@ -143,12 +169,24 @@ impl RangeAllocator {
     }
 
     /// Frees a `range`.
-    #[verifier::external_body]
+    ///
+    /// # Panics
+    ///
+    /// Panics if the freelist has not been initialized by an allocation attempt.
```
rikosellic: This comment does not exist in the source code.

ostd/src/util/range_alloc.rs
```diff
@@ -114,10 +126,24 @@ impl RangeAllocator {
...
+                allocate_range is Some ==> self@.start <= allocate_range->0.start,
+                allocate_range is Some ==> allocate_range->0.end <= self@.end,
+                to_remove is Some ==> allocate_range is Some,
+                to_remove is Some ==> freelist@.contains_key(to_remove->0),
+                to_remove is Some ==> freelist@[to_remove->0].block.end == allocate_range->0.end,
+        )]
         for (key, value) in freelist.iter() {
+            proof! {
+                // `alloc` currently has no callers. Trust that any future caller preserves the
+                // allocator's intended freelist invariant until the lock carries this predicate.
```
rikosellic: Please add `TODO` here.

rikosellic:
Since the `Btree` spec works pretty well, we can contribute it to Verus.

rikosellic:

PR #703, opened 2026-08-12
prove: util::range_alloc with 4 external bodies

Marsman1996:


ostd/src/util/range_alloc.rs
```diff
@@ -2,30 +2,76 @@
...
+#[verus_verify]
 #[derive(Debug)]
 pub struct RangeAllocError;
 
+verus! {
+
+/// Spec model capturing the managed full range of a [`RangeAllocator`].
+pub ghost struct RangeAllocatorModel {
+    pub ghost start: int,
+    pub ghost end: int,
```
rikosellic: There is no need to add `ghost` before the field name, as this is already a ghost struct.

ostd/src/util/range_alloc.rs
```diff
@@ -2,30 +2,76 @@
...
 
+use vstd::prelude::*;
+
 use crate::sync::{PreemptDisabled, SpinLock, SpinLockGuard};
 
+#[verus_verify]
 pub struct RangeAllocator {
     fullrange: Range<usize>,
-    freelist: SpinLock<Option<BTreeMap<usize, FreeRange>>>,
+    freelist: SpinLock<Option<BTreeMap<usize, FreeRange>>, PreemptDisabled>,
```
rikosellic, line 13: Please add a comment here. SpinLock should be improved.

ostd/src/util/range_alloc.rs
Marsman1996: I guess I need to align the skills we used now

rikosellic: BTW, why not use `Range<int>` as the model?

Marsman1996: KVerus now says `Range<int>` is better...

rikosellic:
LGTM. I haven't checked whether the spec for external_body functions are correct though.

rikosellic (approved):

PR #704, opened 2026-08-12
feat: add btree spec in vstd_extra

Marsman1996:


rikosellic:
The `get_mut` part looks good to me, but the Cursor is not very informative. I've been thinking about the model of the Cursor, and what I thought was something like 
```
{
   /// The ordered sequence of keys
   pub keys: Seq<Key>
   /// The position of the cursor. According to the documentation, the cursor is located in the gap between two keys, we can use this field to refer to the index of the element after the cursor.
   pub position: int
  /// The submap of the BtreeMap view that can be accessed by the cursor. Its keys should match the keys field.
  pub map: Map<Key, Value>
}
```

rikosellic:
`borrowed_key_ordering_matches` looks like a good design to me. And it seems this is missing from the spec of `BTreeMap::get` in `vstd`. You may create a PR about this.

verified_libs/vstd_extra/src/external/btree.rs
```diff
@@ -0,0 +1,290 @@
...
+    uninterp spec fn final_map(self) -> Map<Key, Value>;
+}
+
+/// Whether a cursor model consistently represents an ordered map and a gap in that map.
+pub open spec fn cursor_model_wf<Key, Value>(model: CursorMutModel<Key, Value>) -> bool {
+    &&& 0 <= model.position <= model.keys.len()
+    &&& model.keys.no_duplicates()
+    &&& model.keys.to_set() == model.map.dom()
+    &&& increasing_seq(model.keys)
+}
```
rikosellic: Why not make this a method `CursorMutModel::wf`?

verified_libs/vstd_extra/src/external/btree.rs
```diff
@@ -0,0 +1,292 @@
...
+
+/// Whether a borrowed lookup key's ordering agrees with the ordering of stored keys.
+///
+/// This is the semantic requirement imposed on `Key: Borrow<Q>` by the standard library's
+/// borrowed-key `BTreeMap` operations.
+pub uninterp spec fn borrowed_key_ordering_matches<Key, Q: ?Sized>(key: &Q) -> bool;
+
+/// The ordering of a stored key relative to a borrowed lookup key.
+pub uninterp spec fn borrowed_key_cmp<Key, Q: ?Sized>(stored_key: Key, key: &Q) -> Ordering;
```
rikosellic, line 75: Why can we compare these two types？

verified_libs/vstd_extra/src/external/btree.rs
```diff
@@ -0,0 +1,292 @@
...
+        &&& 0 <= self.position <= self.keys.len()
+        &&& self.keys.no_duplicates()
+        &&& self.keys.to_set() == self.map.dom()
+        &&& increasing_seq(self.keys)
+    }
+}
+
+/// Additional abstract and prophetic state for mutable B-tree cursors.
+pub trait CursorMutAdditionalSpecFns<Key, Value>: Sized {
+    spec fn model(self) -> CursorMutModel<Key, Value>;
```
rikosellic: I think we can just name this function as view, and use @ later. If this went into vstd later, we can directly implement `impl View for CursrMut`.

rikosellic (changes requested):

verified_libs/vstd_extra/src/external/btree.rs
```diff
@@ -0,0 +1,292 @@
...
+/// A cursor points at the gap immediately before `keys[position]`. Therefore `peek_next`
+/// accesses `keys[position]`, while `peek_prev` accesses `keys[position - 1]`.
+pub ghost struct CursorMutModel<Key, Value> {
+    /// All keys in the underlying map, in strictly increasing order.
+    pub keys: Seq<Key>,
+    /// The index of the element immediately after the cursor.
+    pub position: int,
+    /// The current contents of the complete map borrowed by the cursor.
+    pub map: Map<Key, Value>,
+}
```
rikosellic, line 40: Is it better to make the fields private and just expose them through methods? Then we can use `#[verifier::type_invariant]` to hide the `wf`.

verified_libs/vstd_extra/src/external/btree.rs
Marsman1996: We do not directly compare `Key` with `Q`. The BTreeMap API requires `Key: Borrow<Q>` and `Q: Ord`, so the stored key is first borrowed as` &Q`, then compared with the lookup key using `Q::cmp`. `borrowed_key_cmp` abstracts `stored_key.borrow().cmp(key)` because vstd does not provide a general specification for `Borrow::borrow`.

verified_libs/vstd_extra/src/external/btree.rs
Marsman1996: `#[verifier::type_invariant]` cannot be used for this purpose because `CursorMutModel` is a spec-level ghost object. Verus type invariants only apply to executable objects and tracked ghost objects; they do not apply to spec objects.

We could still make the fields private and expose accessors, but that would only provide representation hiding. It would not allow us to remove the explicit `cursor@.wf()` guarantees from the specifications. Additionally, `wf()` includes the key-ordering property, which is currently guaranteed only under `obeys_cmp::<Key>()`.

Therefore, KVerus suggests keeping the fields public and retaining the explicit `wf()` postconditions.

rikosellic: Indeed, I forgot about this. Let's just keep these fields public.

verified_libs/vstd_extra/src/external/btree.rs
```diff
@@ -0,0 +1,292 @@
...
+/// Relates a map before and after mutating the value selected by a borrowed key.
+pub open spec fn borrowed_key_mutated<Key, Value, Q: ?Sized>(
+    old_map: Map<Key, Value>,
+    new_map: Map<Key, Value>,
+    key: &Q,
+    old_value: Value,
+    new_value: Value,
+) -> bool {
+    &&& contains_borrowed_key(old_map, key)
+    &&& contains_borrowed_key(new_map, key)
```
rikosellic: `maps_borrowed_key_to_value` already includes `contains_key`, so there is no need to add `contains_borrowd_key` here.

verified_libs/vstd_extra/src/external/btree.rs
```diff
@@ -0,0 +1,292 @@
...
+
+    #[verifier::prophetic]
+    uninterp spec fn final_map(self) -> Map<Key, Value>;
+}
+
+/// Whether a borrowed lookup key's ordering agrees with the ordering of stored keys.
+///
+/// This is the semantic requirement imposed on `Key: Borrow<Q>` by the standard library's
+/// borrowed-key `BTreeMap` operations.
+pub uninterp spec fn borrowed_key_ordering_matches<Key, Q: ?Sized>(key: &Q) -> bool;
```
rikosellic: Is this a type-level specification? The `key` argument does not seem to be very useful. Can we define it as
```
spec fn borrowed_key_ordering_matches<Key: Borrow<Q> + Ord, Q: Ord+?Sized>() -> bool
```

verified_libs/vstd_extra/src/external/btree.rs
```diff
@@ -0,0 +1,288 @@
...
+>[ BTreeMap::<Key, Value, A>::lower_bound_mut::<Q> ](
+    map: &'a mut BTreeMap<Key, Value, A>,
+    bound: Bound<&Q>,
+) -> (cursor: CursorMut<'a, Key, Value, A>)
+    requires
+        match bound {
+            Bound::Included(_) | Bound::Excluded(_) => { borrowed_key_ordering_matches::<Key, Q>()
+            },
+            Bound::Unbounded => true,
+        },
```
rikosellic: We can just use `borrowed_key_ordering_matches` here. There is no need to relax the constraint for `Unbounded`.

verified_libs/vstd_extra/src/external/btree.rs
```diff
@@ -0,0 +1,288 @@
...
+>[ BTreeMap::<Key, Value, A>::upper_bound_mut::<Q> ](
+    map: &'a mut BTreeMap<Key, Value, A>,
+    bound: Bound<&Q>,
+) -> (cursor: CursorMut<'a, Key, Value, A>)
+    requires
+        match bound {
+            Bound::Included(_) | Bound::Excluded(_) => { borrowed_key_ordering_matches::<Key, Q>()
+            },
+            Bound::Unbounded => true,
+        },
```
rikosellic: Same as `lower_bound_mut`.

verified_libs/vstd_extra/src/external/btree.rs
```diff
@@ -0,0 +1,288 @@
...
+            &&& cursor@.map == old(map)@
+            &&& final(map)@ == cursor.final_map()
+            &&& positioned_at_upper_bound(cursor@, bound)
+        },
+;
+
+/// Specification for [`CursorMut::peek_prev`].
+pub assume_specification<'a, 'b, Key, Value, A>[ CursorMut::<'a, Key, Value, A>::peek_prev ](
+    cursor: &'b mut CursorMut<'a, Key, Value, A>,
+) -> (result: Option<(&'b Key, &'b mut Value)>)
```
rikosellic, line 270: Shall we add `old(cursor)@.wf()` as a precondition? The cursor is supposed to be well-formed at all times. Please also update other `peek_next`.

verified_libs/vstd_extra/src/external/btree.rs
```diff
@@ -0,0 +1,288 @@
...
+                &&& *value == old_model.map[*key]
+                &&& new_model.keys == old_model.keys
+                &&& new_model.position == old_model.position
+                &&& new_model.map == old_model.map.insert(*key, *final(value))
+            },
+            None => {
+                &&& old(cursor)@.position == 0
+                &&& final(cursor)@.keys == old(cursor)@.keys
+                &&& final(cursor)@.position == old(cursor)@.position
+                &&& final(cursor)@.map == old(cursor)@.map
```
rikosellic: Just `final(cursor)@==old(cursor)@` should work, same for `peek_next`.

rikosellic:
Apart from existing comments, other codes LGTM.

verified_libs/vstd_extra/src/external/btree.rs
```diff
@@ -0,0 +1,280 @@
...
+    new_value: Value,
+) -> bool {
+    &&& maps_borrowed_key_to_value(old_map, key, old_value)
+    &&& maps_borrowed_key_to_value(new_map, key, new_value)
+    &&& exists|remainder: Map<Key, Value>|
+        {
+            &&& borrowed_key_removed(old_map, remainder, key)
+            &&& borrowed_key_removed(new_map, remainder, key)
+        }
+}
```
rikosellic, line 154: Can you add a broadcast **proof** function for `&Key`, showing that `borrowws_key_mutated` is equivalent to ````old_map.contains_key(key) && old_map[key] == old_value && new_map==old_map.insert(key,new_value)```` ? 
Then there will not be the existential quantifier in this case, and it will make inference easier.

rikosellic:
LGTM, thanks for your contributions! Let's use this to prove `range_alloc` first, then we can contribute this to Verus!

rikosellic (approved):

PR #706, opened 2026-08-12
prove: arch::pci (x86) & io

Marsman1996:
- Add Verus specifications and proofs for x86 PCI configuration-space access.
- Model MMIO and PIO allocation, ownership, and recycling across the OSTD I/O APIs.
- Centralize external `IdAlloc`, x86 port-I/O, and little-endian integer specifications in `vstd_extra::external`.
- Replace the MMIO containment assumption with an explicit boot-window model connected to the concrete allocators by type invariants.

---

Maybe wait until the following PRs are merged:
- [x] https://github.com/asterinas/vostd/pull/742, which proves the id-alloc instead of adding assume


ostd/src/arch/x86/device/io_port.rs
```diff
@@ -7,3 +7,37 @@ pub use x86_64::{
...
+#[verifier::external_body]
+pub struct ExWriteOnlyAccess(WriteOnlyAccess);
+
+/// Trusted specification boundary for values that can be read from an x86 I/O port.
+#[verifier::external_trait_specification]
+pub trait ExPortRead {
+    type ExternalTraitSpecificationFor: PortRead;
+
+    /// A port read can produce any value supplied by the device.
+    unsafe fn read_from_port(port: u16) -> Self where Self: Sized;
```
hiroki-chen: shouldn't we have define specs on IO ports? For instance you cannot read from a serial I/O using invalid port number

Marsman1996: It's because in [the code of `read_from_port` in x86_64 crate](https://github.com/rust-osdev/x86_64/blob/72ff415d2252528042782e5fc3559d74996c5e9a/src/instructions/port.rs#L10-L19), it just accepts any port number without checking.

hiroki-chen: Ok. But this seems worth adding some specs.

Marsman1996: Maybe with more port-ownership capability tokens I guess🤔

hiroki-chen: Yes, or at least we can define some port categories and resolve them to a limited set of numbers available.

Cargo.toml
```diff
@@ -51,3 +51,4 @@ codegen-units = 1
 [workspace.dependencies]
 # Verus
 vstd = { path = "tools/verus/source/vstd", default-features = false, features = ["alloc"] }
+verus_state_machines_macros = { path = "tools/verus/source/state_machines_macros" }
```
rikosellic: Do we need this? I used to believe this is automatiacally included in `vstd`.

Marsman1996: Without this, there will be error like
```
error[E0432]: unresolved import `verus_state_machines_macros`
 --> ostd/src/io/io_port/allocator.rs:3:5
  |
3 | use verus_state_machines_macros::tokenized_state_machine;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^ use of unresolved module or unlinked crate `verus_state_machines_macros`
  |
  = help: if you wanted to use a crate named `verus_state_machines_macros`, use `cargo add verus_state_machines_macros` to add it to your `Cargo.toml`
```

hiroki-chen: Shouldn't this be available directly within the `vstd` crate?

Marsman1996: vstd does not re-export `state_machine!` or `tokenized_state_machine!`

rikosellic: Thanks for your explanation.

hiroki-chen: Remember to add a linebreak at the end of the file

ostd/src/arch/x86/device/io_port.rs
```diff
@@ -7,3 +7,62 @@ pub use x86_64::{
...
+    type ExternalTraitSpecificationFor: PortWrite;
+
+    /// A port write has no modeled logical effect on kernel memory.
+    unsafe fn write_to_port(port: u16, value: Self) where Self: Sized
+        requires
+            valid_io_port_access::<Self>(port as int),
+    ;
+}
+
+} // verus!
```
rikosellic: Please move these assumptions into `vstd_extra/external`.

Marsman1996:
The verification fails caused by the missing spec from https://github.com/asterinas/vostd/pull/702.
Need to wait for it is merged.

verified_libs/vstd_extra/src/external/int_specs.rs
```diff
@@ -27,6 +27,15 @@ pub assume_specification[ u32::is_power_of_two ](self_: u32) -> (r: bool)
...
 
+/// On a little-endian target, converting a native-endian `u32` to little endian is the identity.
+#[cfg(target_endian = "little")]
+pub assume_specification[ u32::to_le ](value: u32) -> (result: u32)
+    returns
+        value,
+    opens_invariants none
+    no_unwind
+;
+
```
rikosellic, line 38: Verus has some specifications about endianness in https://github.com/verus-lang/verus/blob/a0c23e842b787e45332b713a0ff48dc17510a555/source/vstd/endian.rs introduced by https://github.com/verus-lang/verus/pull/2125. I'm not sure whether this will help our verification.

verified_libs/vstd_extra/src/external/io_port.rs
```diff
@@ -0,0 +1,60 @@
...
+};
+
+verus! {
+
+/// Whether `port` is representable in the 16-bit x86 I/O-port address space.
+///
+/// This is only the ISA-level validity condition. It does not claim that a device decodes the
+/// port, that the current CPU context may access it, or that the caller owns it.
+pub open spec fn valid_io_port_number(port: int) -> bool {
+    0 <= port <= u16::MAX as int
```
rikosellic, line 17: This `as int` is unnecessary.

verified_libs/vstd_extra/src/external/io_port.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+/// Whether an access of type `T` fits in the PIO byte range `0..u16::MAX`.
+pub open spec fn valid_io_port_access<T>(port: int) -> bool {
+    &&& valid_io_port_number(port)
+    &&& port + size_of::<T>() <= u16::MAX as int
```
rikosellic, line 23: Again, as int is unnecessary.

rikosellic: Besides, it should be `port + size_of::<T>() <= u16::MAX + 1`, otherwise, you can not read anything from port `0xFFFF`.

rikosellic (changes requested):

verified_libs/vstd_extra/src/external/io_port.rs
rikosellic: Another question, like `BitStore` in `BitVec`, `PortRead` is a public trait that can be implemented by any user. So `port + size_of::<T>() <= u16::MAX` may not be a generally correct requirement. Please take the axiomatized approach,

verified_libs/vstd_extra/src/external/io_port.rs
rikosellic: I find that this definition is used very rarely. Since the `port` argument is `u16` in the method signature, it holds for free. Perhaps we can remove this function?

rikosellic (changes requested):

verified_libs/vstd_extra/src/external/io_port.rs
Marsman1996: The `port + size_of::<T>() <= u16::MAX` is for the `let mut range = port..(port + size_of::<T>() as u16);` in `IoPortAllocator::acquire`, to prevent the overflow.

But after actually there is a potential overflow, and is fixed in the latest Asterinas...

verified_libs/vstd_extra/src/external/io_port.rs
Marsman1996: So in spec mode, are all numbers int？ 🤔

verified_libs/vstd_extra/src/external/int_specs.rs
Marsman1996: not added `assume_specification`, maybe will work in the future

verified_libs/vstd_extra/src/external/io_port.rs
rikosellic: Every number has its original type in spec mode. But arithmetic operations and comparisons have wider support for different types, you can find it here https://verus-lang.github.io/verus/guide/spec-arithmetic.html .

verified_libs/vstd_extra/src/external/io_port.rs
rikosellic: `port + size_of::<T>() <= u16::MAX` makes sense for `u8,u16,u32` that implements `PortRead`. But it does not generally hold for all implementations. The same situation is solved with an `uniterp spec fn obey_bivec_model` in `BitVec`, you can do the same thing here.

rikosellic: > But after actually there is a potential overflow, and is fixed in the latest Asterinas...

Is this code updated to the latest version?

Marsman1996: Nope, the code here is still 0.16.0

Marsman1996: Shall I updated the code here?

Marsman1996: Well, then the port value in Asterinas can never be `0xFFFF`?

rikosellic: If the API doesn't change, we can update the code.

rikosellic: It should be `port + size_of::<T>() <= u16::MAX + 1`, then we can use port `0xFFFF`. If it overflows, that's an Asterinas bug.

Marsman1996:
Update `ostd/src/io/io_port` to 0.18.1 to solve the `port` overflow problem

verified_libs/vstd_extra/src/external/io_port.rs
```diff
@@ -0,0 +1,94 @@
...
+
+/// Trusted: `u32` ports are written and read via `outl`/`inl`.
+pub broadcast axiom fn axiom_u32_pio_model()
+    ensures
+        #[trigger] obeys_pio_model::<u32>(),
+;
+
+/// Under the model, a `T`-access is ISA-valid iff its `size_of::<T>()` bytes fit in
+/// `0..(u16::MAX + 1)`.
+pub broadcast axiom fn axiom_pio_model_access<T>()
```
rikosellic: Should we just add three axioms that  `valid_io_port_access::<$uN>(port) <==> (0 <= port && port + size_of::<T>() <= u16::MAX + 1)`. BTW, will `valid_io_port_access(port: u16)`  be simpler than the current `int` argument? Then we do not need `0 <= port` and `as int` at many places

ostd/src/io/mod.rs
rikosellic, line 1: The change in this file does not seem to match the latest source code.

Marsman1996: Since `io_mem` relies on the latest mm module, only updated the `io_port` related code, and some of the `io_mem` part

ostd/src/io/io_mem/allocator.rs
```diff
@@ -99,34 +187,195 @@ impl IoMemAllocatorBuilder {
...
+        #![trigger registered_io_mem_windows()[m]]
+        0 <= m < registered_io_mem_windows().len() && registered_io_mem_windows()[m].start
+            <= range.start && range.end <= registered_io_mem_windows()[m].end
+}
+
+pub exec static IO_MEM_ALLOCATOR: OnceImpl<IoMemAllocator, TrivialPred>
+    ensures
+        IO_MEM_ALLOCATOR.wf(),
+{
+    OnceImpl::new(Ghost(TrivialPred))
```
rikosellic: `OnceImpl` is our own definition of `Once`. Is it able to assume the specification for `Once`? It is a bit tricky because you may need to add a `Tracked` object to indicate whether an `Once` has been created.

Marsman1996: `OnceImpl::new` is verified and its tracked initialization state is stored internally, so the static initializer can derive `IO_MEM_ALLOCATOR.wf()` from its specification without an additional `Tracked` creation token.

rikosellic:
#769 makes some changes to `OnceImpl` and `AtomicDataWithOwner`, please rebase when it is merged.

ostd/src/io/io_port/allocator.rs
```diff
@@ -1,57 +1,648 @@
...
+    /// Ids claimed by this fragment.
+    pub closed spec fn set(self) -> Set<usize> {
+        self.subset@
+    }
+}
+
+} // verus!
+/// Lock-protected executable bitmap and the state-machine token that models it.
+#[verus_verify]
+struct IoPortAllocatorInner {
```
rikosellic: Not sure what's happening here. Do we need this source code modification?

Marsman1996: We can remove `struct IoPortAllocatorInner` here

verified_libs/vstd_extra/src/external/int_specs.rs
```diff
@@ -27,6 +27,15 @@ pub assume_specification[ u32::is_power_of_two ](self_: u32) -> (r: bool)
     no_unwind
 ;
 
+/// On a little-endian target, converting a native-endian `u32` to little endian is the identity.
+#[cfg(target_endian = "little")]
+pub assume_specification[ u32::to_le ](value: u32) -> (result: u32)
```
rikosellic: We can just write `-> u32` here.

ostd/src/io/io_port/allocator.rs
rikosellic: The code is still using `ModeledIdAlloc`, which is a thin wrapper over `IdAlloc`. Why do we need this?

rikosellic (changes requested):

ostd/src/io/io_port/allocator.rs
Marsman1996: have removed it as well

ostd/src/io/io_mem/mod.rs
```diff
@@ -29,28 +38,70 @@ pub struct IoMem {
...
+    /// Logical offset into the page-aligned mapping.
+    pub closed spec fn offset_spec(&self) -> usize {
+        self.offset
+    }
+}
+
+} // verus!
+#[verus_verify]
 impl HasPaddr for IoMem {
+    #[verus_spec(returns self.paddr_spec())]
```
rikosellic: I believe we can use `verus_verify(dual_spec)` here, so we do not need to define `spec fn` separately. The same is for `length` and `offset`.

Marsman1996: works for `paddr` and `length`, but fail for `offset` (it's private)

rikosellic: `dual_spec` supports a fancy `open/closed/opaque` modifier [recently](https://github.com/verus-lang/verus/pull/2853). So it should work. Please add this in Verus guidline or skill.

Marsman1996: `open`/`closed`/`opaque` controls the generated spec function’s body visibility/opacity
  —not its Rust item visibility. The generated spec clones the executable function’s visibility.

Consequently:

- `#[verus_verify(dual_spec, closed)] fn offset(...)` remains private and cannot appear in public `slice` contracts.
- `#[verus_verify(dual_spec, open)] fn offset(...)` is rejected: Verus requires an `open` function to be `pub`.
- `closed` is already the default for `dual_spec`.

ostd/src/arch/x86/pci.rs
```diff
@@ -37,6 +71,13 @@ pub(crate) fn construct_remappable_msix_address(remapping_index: u32) -> u32 {
 }
 
 /// Encodes the bus, device, and function into a port address for use with the PCI I/O port.
+#[verus_verify]
+#[verus_spec(returns
+    (1u32 << 31)
+        | ((location.bus as u32) << 16)
+        | (((location.device as u32) & 0b11111) << 11)
+        | (((location.function as u32) & 0b111) << 8),
+)]
```
rikosellic: Can we use `(dual_spec,open)` and `returns` here?

ostd/src/io/io_port/mod.rs
```diff
@@ -20,23 +29,183 @@ use crate::{Error, prelude::*};
...
+
+    closed spec fn view(&self) -> u16 {
+        self.port
+    }
+}
+
+impl<T, A> IoPort<T, A> {
+    /// The complete byte range occupied by this typed port lies in the x86 PIO address space.
+    #[verifier::type_invariant]
+    pub open spec fn well_formed(&self) -> bool {
```
rikosellic: Normally, we should name this as `type_inv`, and there is no need to include this in API postconditions.

rikosellic (changes requested):

ostd/src/io/io_port/allocator.rs
```diff
@@ -12,46 +21,532 @@ use crate::{
     sync::{LocalIrqDisabled, SpinLock},
 };
 
+verus! {
+
+/// Identity assigned to the single global PIO allocator during trusted boot initialization.
+pub uninterp spec fn io_port_allocator_instance_id() -> InstanceId;
+
+/// Allocated ids of a bitmap prefix `[0, end)`: `i` is in iff `s[i]`.
+pub(crate) closed spec fn id_alloc_bits(s: Seq<bool>, end: int) -> Set<usize>
```
rikosellic: Use `fold_left`

ostd/src/io/io_port/allocator.rs
```diff
@@ -12,46 +21,532 @@ use crate::{
...
+
+    /// Ids claimed by this fragment.
+    pub closed spec fn set(self) -> Set<usize> {
+        self.subset@
+    }
+}
+
+ghost struct IoPortAllocInvariant;
+
+impl ResourceInvariant<IdAlloc> for IoPortAllocInvariant {
```
rikosellic: Now we have `SimpleResourceInvaraint` when constant is `()`.

ostd/src/io/io_port/allocator.rs
```diff
@@ -12,46 +21,532 @@ use crate::{
...
+    &&& id_alloc_capacity(allocator) == crate::arch::io::MAX_IO_PORT as usize
+}
+
+/// Authority over the set of PIO ids currently allocated by the global allocator.
+///
+/// The `Loc` of `auth` identifies the protocol instance and `auth@` is the set of allocated
+/// ids. [`IoPortClaim`] fragments minted by [`IoPortAllocation::allocate`] transfer the
+/// ownership of an acquired PIO range to the caller.
+pub(super) tracked struct IoPortAllocation {
+    auth: GhostSetAuth<usize>,
```
rikosellic, line 78: Why are we using `usize`, I assume port ids are all `u16`?

rikosellic (changes requested):

ostd/src/io/io_port/allocator.rs
Marsman1996: 1. `IdAlloc`'s verified API is usize
2. The spec `port + size <= u16::MAX + 1` needs it to be larger than `u16`

ostd/src/io/io_port/mod.rs
```diff
@@ -20,23 +29,149 @@ use crate::{Error, prelude::*};
...
+
+    /// Acquires an `IoPort` instance that may overlap with other `IoPort`s.
+    ///
+    /// This method will only mark the first port in the PIO range as occupied.
+    #[verus_spec(result =>
+        with
+            -> claim: Tracked<Option<allocator::IoPortClaim>>,
+        requires
+            size_of::<T>() <= u16::MAX,
+            port as usize + size_of::<T>() <= u16::MAX,
```
rikosellic: This `as usize` seems unnecessary.

ostd/src/io/io_port/allocator.rs
```diff
@@ -12,46 +22,534 @@ use crate::{
...
+}
+
+/// Fragment claiming ownership of a PIO id range handed out by [`IoPortAllocator::acquire`].
+///
+/// A claim asserts that its ids are allocated by the [`IoPortAllocation`] with the same
+/// instance identity, and is consumed when the ids are released by
+/// [`IoPortAllocator::recycle`].
+pub(super) tracked struct IoPortClaim {
```
rikosellic: Do we need this thin wrapper?

ostd/src/io/io_port/allocator.rs
```diff
@@ -12,46 +22,534 @@ use crate::{
...
+    /// otherwise, all ports in the PIO range will be marked as occupied.
+    #[verus_spec(result =>
+        with
+            Tracked(claim_out): Tracked<&mut Tracked<Option<IoPortClaim>>>,
+        requires
+            size_of::<T>() <= u16::MAX,
+            is_overlapping ==> port as usize + size_of::<T>() <= u16::MAX,
+            valid_io_port_access::<T>(port),
+            io_port_allocator_initialized(),
+            (*old(claim_out))@ is None,
```
rikosellic: This definition is very weird. The precondition requires `(*old(claim_out))@ is None`, so why not just give out an `IOPortClaim` and require nothing, like
```
with
        ->   claim_out: Tracked<Option<IoPortClaim>>
```

rikosellic: Qustion2: Can we just record `IOPortClaim` as a field of `IOPort`?

rikosellic (changes requested):

ostd/src/io/io_port/allocator.rs
Marsman1996: Q1: A direct tracked output would be cleaner semantically. However, #[verus_spec] lowers hidden outputs into a tuple return during verification. That makes the existing ? expressions and early return None invalid because the effective return type is no longer Option<_>. The mutable tracked slot preserves the executable signature. Its old(...) is None precondition prevents overwriting an existing linear claim.

Q2: Allocator-created ports could conceptually store Tracked<Option<IoPortClaim>>, but IoPort is also constructed by a const fn for statically reserved ports, which have no allocator claim. The current toolchain cannot const-initialize the tracked None safely; using Tracked::assume_new() would introduce an unjustified trusted token. Also, Verus cannot yet verify the standard Drop implementation, so storing the claim would not currently restore automatic RAII.

A claim field becomes practical if allocated and statically reserved ports are separated into different representations, or once const tracked initialization and Drop verification are supported.

rikosellic: Unfolding `?` is accepatbale, it is a common case in `vostd`.

rikosellic: The RAII rebuttal in Q2 is very weak. The current design requires completely manual manipulation of the tracked token, which is even more disorganized than storing it in `IOPort` and governing it with an invariant. The current design also cannot verify Drop.

For the statically allocated `IOPort` with `const fn new`, one option is to add a `GhostSubsetAuth` argument to the function signature; the other is to store it as `Option<...>`.

ostd/src/io/io_port/allocator.rs
rikosellic: This is reasonable.

ostd/src/io/io_mem/mod.rs
rikosellic: Is this because we need to expose `offset_spec` pulicly?

ostd/src/io/io_mem/allocator.rs
```diff
@@ -60,73 +105,272 @@ impl IoMemAllocator {
...
         Self { allocators }
     }
 
     /// Removes access to a specific memory I/O range.
     ///
     /// All drivers in OSTD must use this method to prevent peripheral drivers from accessing illegal memory I/O range.
+    #[verus_spec(
+        requires
+            range.start < range.end,
+            vstd_extra::panic::may_panic(),
```
Marsman1996, line 177: If we want to add condition for this `may_panic`, we need to further prove the id-alloc module

PR #710, opened 2026-08-17
feat: introduce generic multi-architecture paging/address-space model

DID-Lab-SZU:


ostd/specs/arch/model.rs
```diff
@@ -0,0 +1,63 @@
...
+            0 < Self::max_paddr_spec(),
+            Self::C::BASE_PAGE_SIZE() <= Self::max_paddr_spec(),
+            Self::max_paddr_spec() % Self::C::BASE_PAGE_SIZE() == 0,
+    ;
+}
+
+/// A physical address that can identify a base-page frame for architecture `A`.
+pub open spec fn valid_frame_paddr_for<A: ArchPagingModel>(pa: Paddr) -> bool {
+    pa % A::C::BASE_PAGE_SIZE() == 0 && pa < A::max_paddr_spec()
+}
```
rikosellic, line 29: @Marsman1996  Shall we replace all the original `valid_frame_paddr` with this new definition?

ostd/specs/arch/model.rs
```diff
@@ -0,0 +1,63 @@
...
+                - Self::linear_mapping_base_vaddr_spec(),
+            Self::max_paddr_spec() + Self::linear_mapping_base_vaddr_spec() < usize::MAX,
+    ;
+}
+
+/// Convert a physical address through architecture `A`'s linear mapping.
+pub open spec fn paddr_to_vaddr_for<A: ArchAddressSpaceModel>(pa: Paddr) -> Vaddr {
+    (pa + A::linear_mapping_base_vaddr_spec()) as usize
+}
+
```
rikosellic, line 53: Like above, shall we replace all the original `paddr_to_vaddr_spec` (i.e., `paddr_to_vaddr` in spec mode ) with this new definition?

ostd/specs/arch/model.rs
Marsman1996: I think we should, since the original `valid_frame_paddr` is only for x86 arch 🤔

Marsman1996:
Why we cannot call the copilot now 😣

ostd/specs/arch/model.rs
Marsman1996: It seems they update the `valid_frame_paddr` to call this `valid_frame_paddr_for` now.

rikosellic:
This PR seems to have some overlap with the ongoing https://github.com/asterinas/vostd/pull/675. I prefer to finish that first.

ostd/specs/mm/page_table/owners.rs
```diff
@@ -73,6 +74,183 @@ pub open spec fn vaddr(path: TreePath<NR_ENTRIES>) -> usize {
...
+    } else {
+        path.lemma_index_satisfies_elem_inv(idx);
+        lemma_rec_vaddr_aligned(path, idx + 1);
+        lemma_rec_vaddr_unfold(path, idx);
+
+        let entry_level = (NR_LEVELS - idx) as PagingLevel;
+        let entry_size = page_size(entry_level) as int;
+        lemma_page_size_divides(mapped_level, entry_level);
+        assert(entry_size % mapped_page_size == 0);
+        vstd::arithmetic::div_mod::lemma_fundamental_div_mod(entry_size, mapped_page_size);
```
Marsman1996: Add a grouped import such as
  `use vstd::arithmetic::{div_mod::{lemma_fundamental_div_mod, lemma_mod_multiples_basic, lemma_mod_adds, lemma_mod_equivalence, lemma_small_mod}, mul::{lemma_mul_is_associative, lemma_mul_inequality}};`

Marsman1996 (changes requested):

rikosellic:
Please merge the latest main first, there is some change in lemma names.

dv
rikosellic, line 1: Is this dv updated to the latest version?

ostd/specs/mm/page_table/owners.rs
```diff
@@ -1,30 +1,46 @@
...
     },
 };
 
-use crate::mm::{
-    Paddr, PagingConstsTrait, PagingLevel, Vaddr,
-    frame::meta::{REF_COUNT_MAX, REF_COUNT_UNIQUE, REF_COUNT_UNUSED},
-    page_size,
-    page_table::{EntryOwner, EntryOwnerKind, PageTableEntryTrait, PageTableGuard},
-};
-
```
rikosellic, line 27: `crate::mm` and `crate::specs` xshould be split into two parts.

ostd/src/arch/x86/mm/mod.rs
```diff
@@ -15,19 +14,47 @@ use core::ops::Range;
 pub(crate) use util::{__memcpy_fallible, __memset_fallible};
 //use x86_64::{instructions::tlb, structures::paging::PhysFrame, VirtAddr};
 
-use crate::specs::arch::PAGE_SIZE;
+macro_rules! x86_base_page_size {
+    () => {
+        4096usize
+    };
+}
```
rikosellic, line 21: Is this macro necessary?

rikosellic (changes requested):

rikosellic:
Take a quick look, and it seems good. The role of `CurrentPagingConsts` is still a little ambiguous to me though. To my understanding, it seems to separate the actual constant values from the `PagingConstsTrait`. So we can truly focus on the general part if we only use `PagingConstsTrait`, am I correct? But I notice that `pte_index` uses `CurrentPagingConstsTrait`, is it in an intermediate state?

rikosellic:
Please merge the latest main

ostd/src/mm/page_table/mod.rs
```diff
@@ -892,14 +903,15 @@ fn pte_index<C: PagingConstsTrait>(va: Vaddr, level: PagingLevel) -> (res: usize
 /// This function returns the bit offset of the least significant bit. Take
 /// x86-64 as an example, the `pte_index_bit_offset(2)` should return 21, which
 /// is 12 (the 4KiB in-page offset) plus 9 (index width in the level-1 table).
-fn pte_index_bit_offset<C: PagingConstsTrait>(level: PagingLevel) -> usize
+fn pte_index_bit_offset<C: CurrentPagingConstsTrait>(level: PagingLevel) -> usize
```
rikosellic, line 906: Is this change necessary?

ostd/src/mm/page_table/mod.rs
```diff
@@ -860,7 +870,7 @@ fn nr_pte_index_bits<C: PagingConstsTrait>() -> usize
 }
 
 /// The index of a VA's PTE in a page table node at the given level.
-fn pte_index<C: PagingConstsTrait>(va: Vaddr, level: PagingLevel) -> (res: usize)
+fn pte_index<C: CurrentPagingConstsTrait>(va: Vaddr, level: PagingLevel) -> (res: usize)
```
rikosellic, line 873: Is this change necessary?

ostd/src/mm/page_table/mod.rs
```diff
@@ -179,7 +179,7 @@ pub unsafe trait PageTableConfig: Clone + Debug + Send + Sync + 'static {
     type E: PageTableEntryTrait;
 
     /// The paging constants.
-    type C: PagingConstsTrait;
+    type C: CurrentPagingConstsTrait;
```
rikosellic, line 182: Is this change forever? Or it is in an intermediate state and will go back to `PagingConstsTrait` when we migrate this to different platforms.

DID-Lab-SZU: This is an intermediate step. CurrentPagingConstsTrait extracts the existing equalities with build-selected constants from PagingConstsTrait, since the current page-table structures and proofs still depend on them. Once those structures and proofs are generalized, this bound can return to PagingConstsTrait.

rikosellic: Sounds reasonable, thanks!

rikosellic: I plan to refactor the page table proofs recently, so no need to hurry.

PR #711, opened 2026-08-18
fix: errors caused by latest Verus update

rikosellic:


rikosellic:
@Marsman1996 I've updated to the latest Verus, and it fails on MacOS becuase of rlimit. Please adjust the rlimit for `replace_cur_entry` on MacaOS.

Marsman1996:
> @Marsman1996 I've updated to the latest Verus, and it fails on MacOS becuase of rlimit. Please adjust the rlimit for `replace_cur_entry` on MacaOS.

The rlimit is 200 now, and we still add it?
https://github.com/asterinas/vostd/blob/6c66d2ee3a427795a3ca01610caa3b177ae31389/ostd/src/mm/page_table/cursor/mod.rs#L3992-L3994

rikosellic:
It will be better if you can find a way to fix the proof.

ostd/specs/mm/page_table/cursor/split_while_huge_lemmas.rs
```diff
@@ -588,6 +588,10 @@ impl<C: PageTableConfig> CursorView<C> {
                 vstd::set::lemma_set_choose_len(f);
                 // page_size % new_size == 0
                 assert(m.inv());
+                assert(m.page_size % new_size == 0) by {
```
Marsman1996, line 591: Isn't this @rikosellic just deleted😂

rikosellic: Yeah, do we need this? The only issue is the rlimit of `replace_cur_entry`.

Marsman1996: We need this to pass in the macOS

DID-Lab-SZU: this assertion is necessary in macOS

rikosellic:

PR #715, opened 2026-08-20
Add IRC11 weak AtomicPtr adapter

hiroki-chen:


hiroki-chen:
OK that formatting thing is annoying.

PR #718, opened 2026-08-21
prove: verify utility range operations

DID-Lab-SZU:


ostd/src/util/ops.rs
```diff
@@ -1,28 +1,160 @@
...
+    requires
+        T::obeys_cmp_spec(),
+        T::obeys_partial_cmp_spec(),
+    ensures
+        ret.remaining_seq() == range_difference_spec(*a, *b),
+        forall|i: int|
+            #![auto]
+            0 <= i < ret.remaining_seq().len() ==> !spec_range_is_empty(ret.remaining_seq()[i]),
+{
+    if !(b.start < b.end) {
```
Marsman1996: There are too many exec code changes

Marsman1996 (changes requested):

ostd/src/util/range_alloc.rs
```diff
@@ -135,6 +139,9 @@ impl RangeAllocator {
                 to_remove is Some ==> allocate_range is Some,
                 to_remove is Some ==> freelist@.contains_key(to_remove->0),
                 to_remove is Some ==> freelist@[to_remove->0].block.end == allocate_range->0.end,
+                to_remove is Some ==> size <= freelist@[to_remove->0].block.end,
+                to_remove is Some ==> freelist@[to_remove->0].block.start
+                    <= freelist@[to_remove->0].block.end - size,
```
rikosellic: What's going on in this function, since you add proofs but do not delete any `assume`?

ostd/src/util/ops.rs
```diff
@@ -1,13 +1,345 @@
 // SPDX-License-Identifier: MPL-2.0
+use core::cmp::Ordering;
 use core::ops::Range;
+use vstd::iset::ISet;
```
Marsman1996: better to move these `vstd::` stmts in the top and use a empty line to split to the `core`

verified_libs/vstd_extra/src/external/iter.rs
```diff
@@ -0,0 +1,103 @@
...
+
+        requires
+            iterator_model_valid(self),
+            forall|i: int|
+                0 <= i < iterator_remaining::<Self, <Self as Iterator>::Item>(self).len() ==> (
+                #[trigger] predicate.requires(
+                    (&iterator_remaining::<Self, <Self as Iterator>::Item>(self)[i],),
+                )),
+        ensures
+            iterator_model_valid(filtered),
```
Marsman1996: The postcondition `iterator_model_valid(filtered)` seems too strong. According to the [`Iterator::filter`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter) documentation, `filter` 

> creates an iterator which uses a closure to determine if an element should be yielded

and the closure is evaluated lazily during iteration. Since the closure is an arbitrary `FnMut`, it may not terminate, so the resulting iterator does not necessarily satisfy `will_return_none()`. Could we avoid introducing this termination guarantee here, or require an explicit termination assumption on the predicate?

verified_libs/vstd_extra/src/external/iter.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// `iter.filter(f).next()` is equivalent to `iter.find(f)`.
+    fn filter<P>(self, predicate: P) -> (filtered: Filter<Self, P>) where
+        Self: Iterator + Sized,
+        P: FnMut(&<Self as Iterator>::Item) -> bool,
+
+        requires
+            iterator_model_valid(self),
+            forall|i: int|
+                0 <= i < iterator_remaining::<Self, <Self as Iterator>::Item>(self).len() ==> (
+                #[trigger] predicate.requires(
```
Marsman1996: This requirement appears stronger than the std API contract. The [`Iterator::filter`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter) documentation states that `filter` creates a lazy iterator and that the closure determines whether an element is yielded. The closure is not evaluated when `filter` is called; it is only invoked when the returned iterator is consumed. Therefore, requiring `predicate.requires` for all remaining elements at iterator construction time may unnecessarily restrict valid uses of `filter`.

Marsman1996 (changes requested):
the spec for `filter` seems not right? @rikosellic

verified_libs/vstd_extra/src/external/iter.rs
DID-Lab-SZU: Agreed. I chose to avoid introducing a termination guarantee for `filter`. `iterator_model_valid` no longer implies `will_return_none()`. The concrete array iterator still guarantees `will_return_none()` because it is known to be finite, but a filtered iterator no longer receives that guarantee for an arbitrary `FnMut`.

verified_libs/vstd_extra/src/external/iter.rs
DID-Lab-SZU: I removed the construction-time preconditions from `Iterator::filter`. The source iterator validity and `predicate.requires` conditions are now antecedents of a conditional postcondition. Therefore, constructing a filter remains valid without evaluating or requiring the predicate eagerly; clients only obtain the stronger sequence-model guarantee when they can establish that the predicate is callable for the relevant elements.

rikosellic:
To my knowledge, the Verus team has an [ongoing PR](https://github.com/verus-lang/verus/pull/2515) that adds `Iterator::filter`, and it runs into some subtle technical issues. We should not add our own `filter` specifications. You may skip related proofs for now.

Marsman1996:
> You may skip related proofs for now.

Then there is only the ghost model...

Marsman1996:
~maybe convert to a draft and wait for https://github.com/verus-lang/verus/pull/2515 is merged~

I have merge the PR2515 into [our Verus repo](https://github.com/asterinas/verus) now.

ostd/src/util/ops.rs
```diff
@@ -20,7 +339,44 @@ pub fn range_difference<T: Ord + Copy>(
...
+    let pred = #[verus_spec(keep: bool =>
+        ensures
+            keep == v.start.is_lt(&v.end),
+    )]
+    |v: &Range<T>| !v.is_empty();
+    proof! {
+        assert(iter.obeys_prophetic_iter_laws());
+        assert(iter.decrease() is Some);
+    }
+    let ret = iter.filter(pred);
```
Marsman1996: GPT-5.6-Sol says due to the spec of `Iterator::filter`, we have to add `ret` var here.

Marsman1996 (commented):
To prove 7 lines of exec code, we add 177 lines of proof code, which is  1:25  👀

ostd/src/util/ops.rs
Marsman1996: oops, this is only a comment for @rikosellic

verified_libs/vstd_extra/src/external/cmp.rs
```diff
@@ -0,0 +1,43 @@
...
+        Ordering::Greater => y,
+    }
+}
+
+/// Trusted specification of [`core::cmp::min`].
+pub assume_specification<T: Ord>[ core::cmp::min ](x: T, y: T) -> (res: T)
+    ensures
+        T::obeys_cmp_spec() ==> res == spec_ord_min(x, y),
+;
+
```
rikosellic, line 34: I assume this is already defined in `vstd`? Strange.  https://github.com/verus-lang/verus/blob/b432e82fed7e05090fd53b5e5fc39020f725aabe/source/vstd/std_specs/cmp.rs#L115

Marsman1996:
I changed most `vstd::method` to 
```rust
use vstd::methd;

mehod;
```

However, how about 
```rust
    reveal(vstd::laws_cmp::obeys_partial_cmp_spec_properties);
    reveal(vstd::laws_cmp::obeys_cmp_partial_ord);
    reveal(vstd::laws_cmp::obeys_cmp_ord);
    reveal(vstd::laws_eq::obeys_eq_spec_properties);
```

shall I use the 
`use vstd::laws_cmp`
or just keep it?

@rikosellic

rikosellic:
I think we should use use.

Marsman1996:
> I think we should use use.

Ok, then I will update the guidelines

verified_libs/vstd_extra/src/range.rs
rikosellic, line 1: The file can be moved to vstd_extra.

verified_libs/vstd_extra/src/range.rs
```diff
@@ -0,0 +1,292 @@
...
+        Ordering::Less => x,
+        Ordering::Equal => y,
+        Ordering::Greater => y,
+    }
+}
+
+/// The set denoted by a half-open range.
+pub open spec fn range_as_set<T: PartialOrd>(r: Range<T>) -> ISet<T> {
+    ISet::new(|x: T| r.contains_spec(&x))
+}
```
rikosellic, line 19: I think this is too general. Since we only use number types with a finite range in this project. Why not only consider `T:FiniteRange`, then we can use [`range_len`](https://verus-lang.github.io/verus/verusdoc/vstd/set_lib/trait.FiniteRange.html#tymethod.range_len) or `range_set` directly, enjoying the finiteness property.

verified_libs/vstd_extra/src/range.rs
```diff
@@ -0,0 +1,292 @@
...
+/// The union of the sets denoted by a sequence of ranges.
+pub open spec fn seq_range_union<T: PartialOrd>(s: Seq<Range<T>>) -> ISet<T>
+    decreases s.len(),
+{
+    if s.len() == 0 {
+        ISet::empty()
+    } else {
+        range_as_set(s[0]).union(seq_range_union(s.drop_first()))
+    }
+}
```
rikosellic, line 29: Is this the best way to define the operation with a handwritten recursive function? You can map this to `Set<Set<T>>` (`T:FiniteRange`, of course), then use `flatten`, and reuse lemmas from `vstd`.

rikosellic (changes requested):

verified_libs/vstd_extra/src/range.rs
```diff
@@ -0,0 +1,206 @@
...
+            ) ==> x.partial_cmp_spec(&z) == Some(Ordering::Less),
+        forall|x: T, y: T| x.eq_spec(&y) <==> y.eq_spec(&x),
+        forall|x: T, y: T, z: T| x.eq_spec(&y) && y.eq_spec(&z) ==> x.eq_spec(&z),
+        forall|x: T, y: T|
+            #![trigger x.partial_cmp_spec(&y)]
+            x.is_le(&y) <==> x.cmp_spec(&y) != Ordering::Greater,
+        forall|x: T, y: T| #![trigger x.partial_cmp_spec(&y)] x.is_le(&y) <==> !y.is_lt(&x),
+        forall|x: T|
+            #![trigger x.cmp_spec(&x)]
+            x.is_le(&x) && !x.is_lt(&x) && x.cmp_spec(&x) == Ordering::Equal,
```
rikosellic: I believe these properties have been covered in [`obeys_cmp_spec`](https://verus-lang.github.io/verus/verusdoc/vstd/laws_cmp/fn.obeys_cmp.html)?

verified_libs/vstd_extra/src/range.rs
```diff
@@ -0,0 +1,195 @@
...
+            ),
+        forall|x: T, y: T, z: T|
+            x.partial_cmp_spec(&y) == Some(Ordering::Less) && y.partial_cmp_spec(&z) == Some(
+                Ordering::Less,
+            ) ==> x.partial_cmp_spec(&z) == Some(Ordering::Less),
+        forall|x: T, y: T| x.eq_spec(&y) <==> y.eq_spec(&x),
+        forall|x: T, y: T, z: T| x.eq_spec(&y) && y.eq_spec(&z) ==> x.eq_spec(&z),
+        forall|x: T, y: T|
+            #![trigger x.partial_cmp_spec(&y)]
+            x.is_le(&y) <==> x.cmp_spec(&y) != Ordering::Greater,
```
rikosellic: Actually, I mean all the post conditions are covered by `obeys_cmp_spec`.

verified_libs/vstd_extra/src/range.rs
```diff
@@ -0,0 +1,194 @@
...
+    } else {
+        Seq::empty()
+    }
+}
+
+proof fn lemma_seq_range_union_contains<T: FiniteRange>(s: Seq<Range<T>>, x: T)
+    ensures
+        seq_range_union(s).contains(x) <==> exists|i: int|
+            #![trigger range_as_set(s[i]).contains(x)]
+            0 <= i < s.len() && range_as_set(s[i]).contains(x),
```
rikosellic: `Seq` has a `Seq::any` method that encapsulates this `exists`, that may be cleaner.

ostd/src/util/ops.rs
```diff
@@ -8,19 +25,104 @@ use core::ops::Range;
...
+    proof! {
+        reveal(obeys_cmp_partial_ord);
+        reveal(obeys_cmp_ord);
+    }
     let r = if b.is_empty() {
         [a.clone(), b.clone()]
     } else {
-        [a.start..min(a.end, b.start), max(a.start, b.end)..a.end]
+        // `Ord::{min,max}` are the specified equivalents of `core::cmp::{min,max}`.
+        [a.start..a.end.min(b.start), a.start.max(b.end)..a.end]
```
rikosellic: You mayaadd an `assume_specification` for `min` and `max`.

ostd/src/util/ops.rs
```diff
@@ -8,19 +25,104 @@ use core::ops::Range;
...
+    requires
+        obeys_cmp::<T>(),
+        finite_range_matches_ord::<T>(),
+    ensures
+        ret.obeys_prophetic_iter_laws() && ret.will_return_none() ==> {
+            &&& ret.remaining() == range_difference_seq(*a, *b)
+            &&& ret.remaining().len() <= 2
+            &&& forall|i: int|
+                0 <= i < ret.remaining().len() ==> (
+                #[trigger] ret.remaining()[i]).start.is_lt(&ret.remaining()[i].end)
```
rikosellic: This forall clause can be expressed by `Seq::all`.

verified_libs/vstd_extra/src/range.rs
```diff
@@ -0,0 +1,194 @@
...
+        Ordering::Less => x,
+        Ordering::Equal => y,
+        Ordering::Greater => y,
+    }
+}
+
+/// The finite set denoted by a half-open range.
+pub open spec fn range_as_set<T: FiniteRange>(r: Range<T>) -> Set<T> {
+    T::range_set(r.start, r.end)
+}
```
rikosellic, line 19: I prefer this function to be defined as an associated fucntion like 
```
spec fn view_set(self) -> Set<T>
```
Since `range` is not defined in `vstd_extra`, you may have to define a helper trait to achieve this. You can find the example in `auxiliary` where I define a `tracked_borrow_mut` method for `Option`. It is simpler in this case since this is only a spec function.

verified_libs/vstd_extra/src/range.rs
```diff
@@ -0,0 +1,194 @@
...
+        obeys_cmp::<T>(),
+        0 <= i < range_difference_seq(a, b).len() - 1,
+    ensures
+        range_difference_seq(a, b)[i].end.is_le(&range_difference_seq(a, b)[i + 1].start),
+{
+    reveal(obeys_partial_cmp_spec_properties);
+    reveal(obeys_cmp_partial_ord);
+    reveal(obeys_cmp_ord);
+    reveal(obeys_eq_spec_properties);
+}
```
rikosellic, line 57: This function is ugly and only used as an internal proof in `ops`. Perhaps you can remove this?

ostd/src/util/ops.rs
rikosellic: Maybe the upstream Verus PR is not mature enough, we may find a way to fix this later. LGTM currently.

verified_libs/vstd_extra/src/range.rs
```diff
@@ -0,0 +1,194 @@
...
+        0 <= i < s.len() && range_as_set(s[i]).contains(x) {
+        let i = choose|i: int|
+            #![trigger range_as_set(s[i]).contains(x)]
+            0 <= i < s.len() && range_as_set(s[i]).contains(x);
+        assert(range_sets.to_set().contains(range_sets[i]));
+        range_sets.to_set().lemma_flatten_contains(x);
+    }
+}
+
+proof fn lemma_seq_range_union_small<T: FiniteRange>(s: Seq<Range<T>>)
```
rikosellic: This function is used only once, you can inline it in the following proof.

rikosellic (changes requested):

Marsman1996:
@rikosellic  
If we add a `filter_with_spec` wrapper in vstd_extra, then the code change from
```rust
    let iter = r.into_iter();
    let pred = #[verus_spec(keep: bool =>
        ensures
            keep == v.start.is_lt(&v.end),
    )]
    |v: &Range<T>| !v.is_empty();
    let ret = iter.filter(pred);
    ret
```
to
```rust
    #[verus_spec(with Ghost(|v: Range<T>| v.start.is_lt(&v.end)))]
    vstd_extra::external::iter::filter_with_spec(
        r.into_iter(),
        #[verus_spec(keep: bool =>
            ensures
                keep == v.start.is_lt(&v.end),
        )]
        |v: &Range<T>| !v.is_empty(),
    )
```
which is more similar to the original code
```rust
r.into_iter().filter(|v| !v.is_empty())
```
Is this an acceptable change?

rikosellic:
I think the current design is good enough. We can just wait the improvement from Verus.

Marsman1996:
@Je5s1e 
merge the latest commit to solve the assert failure.
run `make clean` to solve the `rlimit` error

Je5s1e:
您好，邮件我已收到，辛苦您了！我会尽快阅读并且给您回复。如有急事，请直接致电：15207340242

verified_libs/vstd_extra/src/external/iter.rs
```diff
@@ -0,0 +1,35 @@
...
+//! Specification for owned-array iteration not yet modeled by `vstd`.
+//!
+//! The array iterator contract follows Rust 1.97.1's
+//! `library/core/src/array/iter.rs` implementation, which initializes an
+//! iterator over the array in left-to-right order.
+//!
+//! # Verified Properties
+//!
+//! The trusted contract records the iterator's element order, termination,
+//! remaining length, and prophetic iterator laws.
```
rikosellic: This part is unnecessary. Typically, we only add this for APIs.

verified_libs/vstd_extra/src/range.rs
```diff
@@ -0,0 +1,205 @@
...
+}
+
+/// Whether the finite-range model agrees with the ordering model.
+pub open spec fn finite_range_matches_ord<T: FiniteRange + Ord>() -> bool {
+    forall|x: T, lo: T, hi: T| T::in_range(x, lo, hi) <==> lo.is_le(&x) && x.is_lt(&hi)
+}
+
+/// Operational spec of `range_difference`: the two candidate ranges,
+/// dropping any empty ones, matching the executable control flow.
+pub open spec fn range_difference_seq<T: Ord>(a: Range<T>, b: Range<T>) -> Seq<Range<T>> {
```
rikosellic: `range_difference_seq` and `lemma_range_dofferent_set` can be moved to `util/ops.rs` (no need to create a new folder in `specs` since they are small). It is the spec for `range_difference`. BTW, `range_difference_spec` may be a better name.

verified_libs/vstd_extra/src/range.rs
```diff
@@ -0,0 +1,205 @@
...
+}
+
+/// The union of the sets denoted by a sequence of ranges.
+pub open spec fn seq_range_union<T: FiniteRange>(s: Seq<Range<T>>) -> Set<T> {
+    s.map_values(|r: Range<T>| r.view_set()).to_set().flatten()
+}
+
+closed spec fn range_contains<T: FiniteRange>(x: T) -> spec_fn(Range<T>) -> bool {
+    |r: Range<T>| r.view_set().contains(x)
+}
```
rikosellic: Do we actually need this function? Will inlining this funtion be easier?

verified_libs/vstd_extra/src/range.rs
```diff
@@ -0,0 +1,205 @@
...
+
+    reveal(obeys_partial_cmp_spec_properties);
+    reveal(obeys_cmp_partial_ord);
+    reveal(obeys_cmp_ord);
+    reveal(obeys_eq_spec_properties);
+    let s = range_difference_seq(a, b);
+    if s.len() == 0 {
+        assert forall|x: T| !seq_range_union(s).contains(x) by {
+            lemma_seq_range_union_contains(s, x);
+            reveal(Seq::any);
```
rikosellic: `Seq::any` is an `open spec fn`, so I assume you don't need to reveal it.

ostd/src/util/ops.rs
```diff
@@ -8,19 +27,111 @@ use core::ops::Range;
...
+    proof! {
+        reveal(obeys_cmp_partial_ord);
+        reveal(obeys_cmp_ord);
+    }
     let r = if b.is_empty() {
         [a.clone(), b.clone()]
     } else {
-        [a.start..min(a.end, b.start), max(a.start, b.end)..a.end]
+        // `Ord::{min,max}` are the specified equivalents of `core::cmp::{min,max}`.
+        [a.start..a.end.min(b.start), a.start.max(b.end)..a.end]
```
rikosellic: I remember I requested to assume specifications for `core::cmp::{min, max}`.

rikosellic (changes requested):

ostd/src/util/ops.rs
Marsman1996: image

verified_libs/vstd_extra/src/range.rs
```diff
@@ -0,0 +1,67 @@
...
+///
+/// # Preconditions
+///
+/// No additional preconditions are required.
+///
+/// # Postconditions
+///
+/// An element belongs to the union if and only if it belongs to at least one
+/// range in the sequence.
+pub proof fn lemma_seq_range_union_contains<T: FiniteRange>(s: Seq<Range<T>>, x: T)
```
rikosellic, line 33: I think we only need lines 39-40. Other comments are unnecessary. @Marsman1996 Maybe our guideline is too strict?

ostd/src/util/ops.rs
```diff
@@ -1,26 +1,209 @@
...
+                reveal(obeys_cmp_ord);
+                reveal(obeys_eq_spec_properties);
+            }
+            assert(seq_range_union(ret.remaining()) ==
+                (*a).view_set().difference((*b).view_set())) by {
+                lemma_range_difference_set(*a, *b);
+            }
+        }
+    }
+    ret
```
rikosellic: I suppose all these functions are open, and they are not hidden using a `hide` command. Is this caused by the guideline? @Marsman1996

ostd/src/util/ops.rs
```diff
@@ -1,26 +1,209 @@
...
+        finite_range_matches_ord::<T>(),
+    ensures
+        seq_range_union(range_difference_spec(a, b)) == a.view_set().difference(b.view_set()),
+{
+    broadcast use range_set_properties;
+
+    reveal(obeys_partial_cmp_spec_properties);
+    reveal(obeys_cmp_partial_ord);
+    reveal(obeys_cmp_ord);
+    reveal(obeys_eq_spec_properties);
```
rikosellic, line 61: These reveals seems to be redundant.

rikosellic (changes requested):

verified_libs/vstd_extra/src/range.rs
rikosellic: For a simple proof lemma like this, we only need a short description. The verified properties, pre-condition and post-condition sections are only for APIs.

Marsman1996: LLM always write more comments even if we currently require `Avoid restating code, speculating about tool limitations, or recounting failed proof attempts.`

Maybe a more restrict guideline is needed

ostd/src/util/ops.rs
Marsman1996: Not sure, maybe related to the `organize-proof-imports` guideline, and it uses a `reveal` as an example.

However, I have not met the situation that the LLM uses this large amount of `reveal` stmt.🤔

verified_libs/vstd_extra/src/external/range.rs
```diff
@@ -45,6 +49,17 @@ pub fn range_usize_is_empty(r: &Range<usize>) -> (ret: bool)
     !(r.start < r.end)
 }
 
+/// Trusted specification of `Range::is_empty` from Rust 1.97.1's
```
Marsman1996: It's 1.98.0 now, and we do not need this detail comment

verified_libs/vstd_extra/src/external/iter.rs
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: MPL-2.0
+//! Specification for owned-array iteration not yet modeled by `vstd`.
+//!
+//! The array iterator contract follows Rust 1.97.1's
```
Marsman1996: Same here

ostd/src/util/ops.rs
```diff
@@ -1,26 +1,205 @@
...
+)]
+pub fn range_difference<T: Ord + Copy + FiniteRange>(
     a: &Range<T>,
     b: &Range<T>,
 ) -> impl Iterator<Item = Range<T>> {
     use core::cmp::{max, min};
 
+    proof! {
+        reveal(obeys_cmp_partial_ord);
+        reveal(obeys_cmp_ord);
```
rikosellic, line 112: Please review all `reveal`s, they all seems redundant to me. @Marsman1996 Should we add this in coding guidline?

rikosellic (changes requested):

verified_libs/vstd_extra/src/external/cmp.rs
```diff
@@ -0,0 +1,46 @@
...
+/// Returns `x` when `y` compares less than it, and returns `y` otherwise.
+pub open spec fn spec_ord_max<T: Ord>(x: T, y: T) -> T {
+    match y.cmp_spec(&x) {
+        Ordering::Less => x,
+        Ordering::Equal => y,
+        Ordering::Greater => y,
+    }
+}
+
+/// Returns the minimum, choosing the first argument when they compare equal.
```
Marsman1996: It seems these are not needed, the review says:
external/cmp.rs duplicates vstd `Ord::min/max`

Marsman1996 (changes requested):

ostd/src/util/ops.rs
Marsman1996: will add tomorrow

verified_libs/vstd_extra/src/external/cmp.rs
Marsman1996: @Je5s1e Are you using the GPT-6 to review?

ostd/src/util/ops.rs
```diff
@@ -1,26 +1,205 @@
...
+            seq![left]
+        }
+    } else if right.start.is_lt(&right.end) {
+        seq![right]
+    } else {
+        Seq::empty()
+    }
+}
+
+/// Proves that [`range_difference_spec`] denotes the finite set difference `a - b`.
```
Marsman1996: No need for these comments.

ostd/src/util/ops.rs
```diff
@@ -1,26 +1,205 @@
...
+            assert(pred(s[1]));
+        }
+    }
+    assert forall|x: T|
+        #![trigger a.view_set().contains(x)]
+        seq_range_union(s).contains(x) <==> a.view_set().difference(b.view_set()).contains(x) by {}
+}
+
+} // verus!
 /// Calculates the [difference] of two [`Range`]s, i.e., `a - b`.
```
Marsman1996, line 79: Chnage line here

ostd/src/util/ops.rs
```diff
@@ -1,26 +1,205 @@
...
 ///
 /// This method will return 0, 1, or 2 ranges. All returned ranges are
 /// guaranteed to be non-empty and non-overlapping. The returned ranges
 /// will be sorted in ascending order.
 ///
 /// [difference]: https://en.wikipedia.org/wiki/Set_(mathematics)#Set_difference
-pub fn range_difference<T: Ord + Copy>(
+///
```
Marsman1996: No need for all of these comments, since `range_difference` is not an API

Marsman1996 (changes requested):

ostd/src/util/ops.rs
Marsman1996: Damn, the verusfmt makes this not work

ostd/src/util/ops.rs
DID-Lab-SZU: the 3 reveals here cant be removed, as remove them cant make

ostd/src/util/ops.rs
rikosellic: @DID-Lab-SZU I'm still seeing these reveals. Can we remove all of them?

ostd/src/util/ops.rs
```diff
@@ -1,26 +1,159 @@
...
+        ret.obeys_prophetic_iter_laws() && ret.will_return_none() ==> {
+            &&& ret.remaining() == range_difference_spec(*a, *b)
+            &&& ret.remaining().len() <= 2
+            &&& ret.remaining().all(
+                |range: Range<T>| range.start.is_lt(&range.end),
+            )
+            &&& forall|i: int|
+                0 <= i < ret.remaining().len() - 1 ==> (
+                #[trigger] ret.remaining()[i]).end.is_le(&ret.remaining()[i + 1].start)
+            &&& seq_range_union(ret.remaining()) == (*a).view_set().difference((*b).view_set())
```
rikosellic: The dereference operator in `*a` and `*b` may be unnecessary.

rikosellic (changes requested):

ostd/src/util/ops.rs
DID-Lab-SZU: the 3 reveals here cant be removed, as remove them cant make

rikosellic: Ah I see, they are opaque.

ostd/src/util/ops.rs
rikosellic: I find these functions to be opaque. It looks good.

ostd/src/util/ops.rs
```diff
@@ -1,26 +1,159 @@
...
+        RangeExtraFns, finite_range_matches_ord, lemma_seq_range_union_contains, seq_range_union,
+    },
+};
+
 use core::ops::Range;
 
+verus! {
+
+/// Operational spec of `range_difference`: the two candidate ranges,
+/// dropping any empty ones, matching the executable control flow.
```
rikosellic: I think /// Operational spec of `range_difference`  is enough.

rikosellic:

rikosellic (approved):

Marsman1996 (approved):
LGTM!

PR #723, opened 2026-08-24
Add tracked RCU read pools

hiroki-chen:
An RCU reader obtains only one fraction of the read lease so that we need to bookkeep who is borrowing the reader to ensure that we cannot accidentally drop it. We can restore the full complete ownership of that object only until no active lease exists.

So this PR defines the spec and proofs needed to track this fact. This does not contain any changes to the exec code and is pure spec-level addition.

### DIFF
--- a/verified_libs/vstd_extra/src/rcu_read_lease.rs
+++ b/verified_libs/vstd_extra/src/rcu_read_lease.rs
@@ -1,0 +1,915 @@
+//! Unbounded fractional read leases for delayed reclamation.
+//!
+//! An owner-side accumulator stores one linear resource in Verus' Leaf-style
+//! storage protocol. Each reader-held lease receives half of the accumulator's
+//! current rational fraction, so the number of outstanding leases has no fixed
+//! integer bound. Reclamation can recover the resource only after all leases
+//! have been returned and the accumulator fraction is whole.
+use vstd::{
+    prelude::*,
+    resource::{Loc, frac_opt::Frac},
+};
+
+verus! {
+
+/// Owner-side fractional accumulator for one delayed-reclamation resource.
+pub tracked struct RcuLeaseAccumulator<T> {
+    frac: Frac<T>,
+}
+
+/// Reader-held fractional permission split from an [`RcuLeaseAccumulator`].
+pub tracked struct RcuReadLease<T> {
+    frac: Frac<T>,
+}
+
+/// Reader-held lease registered under an allocation key and a unique lease ID.
+///
+/// The private `lease_id` names the matching [`RcuActiveReadLeaseRecord`] in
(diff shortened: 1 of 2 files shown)


rikosellic:
The resource names are a bit confusing. It's hard to distinguish different RCU resources. Can you name them more informatively? Or are these names defined in the related papers' artifacts?

hiroki-chen:
These names are not inherited from the paper artifacts; they represent implementation-specific resources connecting abstract RCU protection to fractional physical permissions. We renamed them according to their ownership roles and removed the unusedsimplified registry. This currently avoids a collision with vostd's existing RcuReadPool alias.

verified_libs/vstd_extra/src/rcu_read_lease.rs
```diff
@@ -0,0 +1,914 @@
...
+        self.lease.fraction()
+    }
+
+    pub proof fn borrow(tracked &self) -> (tracked resource: &T)
+        ensures
+            *resource == self.resource(),
+    {
+        self.lease.borrow()
+    }
+}
```
rikosellic, line 109: Is it better to name it as `tracked_borrow`?

verified_libs/vstd_extra/src/rcu_read_lease.rs
```diff
@@ -0,0 +1,914 @@
...
+
+/// Owner-side fractional accumulator for one delayed-reclamation resource.
+pub tracked struct RcuLeaseAccumulator<T> {
+    frac: Frac<T>,
+}
+
+/// Reader-held fractional permission split from an [`RcuLeaseAccumulator`].
+pub tracked struct RcuReadLease<T> {
+    frac: Frac<T>,
+}
```
rikosellic, line 23: Will `RcuReadLease` be used on its own?

verified_libs/vstd_extra/src/rcu_read_lease.rs
hiroki-chen: Because they are purely verus variables so either `tracked_borrow` or `borrow` works. If we need to make the codebase consistent then I think `tracked_borrow` should be better.

rikosellic: Then we should take `tracked_borrow` to be more consistent.

verified_libs/vstd_extra/src/rcu_read_lease.rs
hiroki-chen: Yes. The thing is that initially I think there may be some use cases where we need to extend `Frac` so I use a wrapper instead of a type alias to `Frac`; for this PR read lease acts in the same way as `Frac` does.

PR #728, opened 2026-08-31
prove: model range allocator freelist initialization

Marsman1996:
finish the prove with the new Spinlock PR.
have to modify the exec code for `drop` and verification.


ostd/src/util/range_alloc.rs
```diff
@@ -1,21 +1,112 @@
...
+tracked struct FreelistResource {
+    initialized_auth: GhostSetAuth<()>,
+    initialized_empty: GhostSubset<()>,
+    initialized_witness: Option<GhostPersistentSingleton<()>>,
+}
+
+struct FreelistConstant {
+    fullrange: Range<int>,
+    initialized_id: Loc,
+}
```
rikosellic, line 22: Can we mark `FreelistConstant` and `Freelist Invariant` as `ghost struct`?

ostd/src/util/range_alloc.rs
```diff
@@ -1,21 +1,112 @@
...
+    fullrange: Range<int>,
+    freelist: Map<usize, FreeRange>,
+    key: usize,
+)
+    requires
+        freelist_wf(fullrange, freelist),
+    ensures
+        freelist_wf(fullrange, freelist.remove(key)),
+{
+}
```
rikosellic: This lemma seems very trivial, can we remove this?

ostd/src/util/range_alloc.rs
```diff
@@ -1,21 +1,112 @@
...
+    key: usize,
+    value: FreeRange,
+)
+    requires
+        freelist_wf(fullrange, freelist),
+        fullrange.start <= value.block.start <= value.block.end <= fullrange.end,
+    ensures
+        freelist_wf(fullrange, freelist.insert(key, value)),
+{
+}
```
rikosellic: Can we also remove this?

ostd/src/util/range_alloc.rs
```diff
@@ -1,21 +1,112 @@
...
 
 use crate::sync::{PreemptDisabled, SpinLock, SpinLockGuard};
 
+verus! {
+
+tracked struct FreelistResource {
+    initialized_auth: GhostSetAuth<()>,
+    initialized_empty: GhostSubset<()>,
+    initialized_witness: Option<GhostPersistentSingleton<()>>,
+}
```
rikosellic: These designs are correct, but a bit ugly, can you use `Token` in `vstd_extra::resource` or add some type aliases in this file?

ostd/src/util/range_alloc.rs
Marsman1996: So for all newly added `struct`s, we shall try to mark them as ghost as much as possible?

rikosellic: Indeed

ostd/src/util/range_alloc.rs
```diff
@@ -79,6 +204,17 @@ impl RangeAllocator {
...
+                target_node is Some ==> freelist@.contains_key(target_node->0),
+                target_node is Some ==>
+                    freelist@[target_node->0].block.start <= allocate_range.start,
+                target_node is Some ==>
+                    allocate_range.end <= freelist@[target_node->0].block.end,
+                target_node is Some ==>
+                    left_length == allocate_range.start
+                        - freelist@[target_node->0].block.start,
+                target_node is Some ==>
+                    right_length == freelist@[target_node->0].block.end - allocate_range.end,
```
rikosellic: I prefer the following styling:
```
<return_value_name> matches Some(<var_name>) ==> {
 &&& ...
 &&& ...
}
```

ostd/src/util/range_alloc.rs
Marsman1996: `vstd_extra::resource::Token<N>` is not a sound replacement here: it is a bounded fractional token and cannot be duplicated indefinitely, while every allocation attempt must be able to return a persistent initialization witness. The current `GhostPersistentSingleton` is the right algebra.

rikosellic: How about `GhostSetAuth` and `GhostSubset`?

Marsman1996: use type aliases

rikosellic:
I'm thinking whether you can encapsulate the GhostSetAuth<()> and PersistentSet in a data structure and add it in vstd_extra::resource. It is a special case where we only care about the existence of a token instead of its content, so we do not need to expose the () implementation. It may be useful in other places.

Marsman1996:
> I'm thinking whether you can encapsulate the GhostSetAuth<()> and PersistentSet in a data structure and add it in vstd_extra::resource. It is a special case where we only care about the existence of a token instead of its content, so we do not need to expose the () implementation. It may be useful in other places.

Updated

verified_libs/vstd_extra/src/resource/mod.rs
```diff
@@ -1,3 +1,4 @@
 //! Resource algebras based on Iris PCM and Leaf storage-protocol.
 pub mod ghost_resource;
+pub mod persistent_flag;
```
rikosellic: Maybe `flags` instead of `persistent_flags`.

ostd/src/util/range_alloc.rs
```diff
@@ -1,21 +1,80 @@
...
 };
 
 use alloc::collections::btree_map::BTreeMap;
 use core::ops::Range;
 
 use crate::sync::{PreemptDisabled, SpinLock, SpinLockGuard};
 
+verus! {
+
+pub type FreelistInitializationToken = GhostPersistentFlag;
```
rikosellic: This type alias is no longer necessary.

ostd/src/util/range_alloc.rs
```diff
@@ -1,21 +1,80 @@
...
+
+tracked struct FreelistResource {
+    initialized_auth: GhostFlagAuth,
+    initialized_witness: Option<FreelistInitializationToken>,
```
rikosellic: Do we even need to always record `GhostFlagAuth`? Can we change the interface of `GhostFlagAuth::set` to
```
 proof fn set(tracked self) -> (tracked result: GhostPersistentFlag)
```
so that it will consume the `Auth` to produce a ` GhostPersistentFlag`. Then we can duplicate `GhostPersistentFlag` everywhere. The `FreeListResource` can accordingly be simplified to `Sum<GhostFlagAuth,GhostPersistentFlag>`. You can just define it in `ResourceInvariant`, the wrapper type is no longer required.

And you can give more informative names, like `OneShotPending` and `OneShotSet`, or other better names?

ostd/src/util/range_alloc.rs
rikosellic, line 1: Do we need to give out a `OneShotSet` to the callers? It does not seem necessary.

ostd/src/util/range_alloc.rs
```diff
@@ -1,21 +1,66 @@
...
+            Sum::Left(pending) => {
+                &&& pending.id() == constant.initialized_id
+                &&& freelist is None
+            },
+            Sum::Right(set) => {
+                &&& set.id() == constant.initialized_id
+                &&& freelist is Some
+            },
+        }
+        &&& freelist is Some ==> freelist_wf(constant.fullrange, freelist->0@)
```
rikosellic: This condidition can be put into the Right block.

ostd/src/util/range_alloc.rs
```diff
@@ -213,22 +328,52 @@ impl RangeAllocator {
...
+        proof! {
+            if *resource is Right {
+                initialized = resource.tracked_borrow_right().duplicate();
+            } else {
+                let tracked replacement = OneShotPending::new();
+                let tracked pending = resource.tracked_swap_left(replacement);
+                let tracked set = pending.set();
+                initialized = set.duplicate();
+                *resource = Sum::Right(set);
+            }
```
rikosellic: Not sure why we need a placeholder. If the `Sum` API is not sufficient, please add one.

ostd/src/util/range_alloc.rs
Marsman1996: We need this, otherwise, `free` cannot distinguish initialized from uninitialized state through `&self`

rikosellic: Indeed, it makes sense.

ostd/src/util/range_alloc.rs
```diff
@@ -158,25 +251,46 @@ impl RangeAllocator {
...
+        };
+        lock_guard.drop();
+        #[verus_spec(with |= Tracked(initialized))]
+        res
     }
 
     /// Frees a `range`.
     #[verus_spec(
+        with
+            Tracked(initialized): Tracked<Option<OneShotSet>>,
```
rikosellic: Just requiring `Tracked<OneShot>` may be simpler, the `initialized is None ==> may_panic(),` is not very meaningful in this case.

Marsman1996: ~~If that's the case, then we do not need to give out a `OneShotSet` to the callers.~~
made a mistake above, never mind.

However, how to justify a spec is meaningful or not 🤔

rikosellic: In essence, this is just a taste issue. To justify a bit, `may_panic` is preferred to indicate that some valid value of the exec input parameter can still cause a panic, for example, an overflow in `Segment::slice`. `OneShotSet` however, is a ghost object that is created in verification to ensure the APIs are called in the correct order. By requiring a `OneShotSet`, one must have called `alloc` first before he calls `free`. This is different from the `Segment::Slice` case, and is closer to `Frame::into_raw` and `Frame::from_raw`.

rikosellic:
I've adjusted some Resource APIs to make the coding style more consistent with other resources. LGTM now, thanks for your contributions!

rikosellic (approved):

PR #737, opened 2026-09-02
Mark one-shot flags objective under IRC11

hiroki-chen:


rikosellic:
Just curious, what does objective mean?

hiroki-chen:
> Just curious, what does objective mean?

In weak memory model, we have "objective" facts globally. For example, you can say that for an atomic location $L$, some invariants hold. However, each thread sees only part of the truth so what a thread knows is "subjective". This trait is to distinguish between these two concepts.

rikosellic:
Ah, I see. This concept is indeed mentioned in the paper. I need to learn more about RC11. Thanks for your explanation!

hiroki-chen:
Wait for @rikosellic to fix this
```
WARNING: (8628,1): pattern does not contain all quantified variables.
error: assertion failed
   --> ostd/src/../specs/sync/abstract_lock.rs:351:53
    |
351 |             && s.inv_not_locked_iff_no_cs() implies s_prime.inv_not_locked_iff_no_cs() by {
    |                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ assertion failed

verification results:: 1529 verified, 1 errors
warning: `ostd` (lib) generated 443 warnings (49 duplicates)
error: could not compile `ostd` (lib) due to 1 previous error; 444 warnings emitted
Verification failed for ostd
```

Marsman1996:
So this PR is included in #740

PR #738, opened 2026-09-03
sync(sync): update to 0.18.1

Marsman1996:
Update the sync code to Asterinas 0.18.1
Shall we add a version note just like in [mainline branch](https://github.com/asterinas/vostd/tree/phaseII/mainline)?

### DIFF
--- a/ostd/src/sync/rwlock.rs
+++ b/ostd/src/sync/rwlock.rs
@@ -478,8 +478,8 @@ impl<T  /*: ?Sized*/ , G: SpinGuardian> RwLock<T, G> {
         if lock & (WRITER | MAX_READER | BEING_UPGRADED) == 0 {
             Some(
                 RwLockReadGuard {
-                    inner: self,
                     guard,
+                    inner: self,
                     tracked_token: Tracked(read_token.tracked_unwrap()),
                 },
             )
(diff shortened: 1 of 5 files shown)


ostd/src/sync/rwlock.rs
```diff
@@ -617,26 +617,25 @@ impl<T  /*: ?Sized*/ , G: SpinGuardian> RwLock<T, G> {
...
+*/
     /// Returns a raw pointer to the underlying data.
     ///
     /// This method is safe, but it's up to the caller to ensure that access to the data behind it
     /// is still safe.
+    #[verifier::external_body]
     pub(super) fn as_ptr(&self) -> *mut T {
-        self.val.get()
+        // self.val.get()
+        core::ptr::from_ref(&self.val) as *const T as *mut T
```
rikosellic: This change does not match the Asterinas source code. Just comment it.

rikosellic (changes requested):

rikosellic:
Yeah it will be good to add a version note.

ostd/src/sync/rwlock.rs
Marsman1996: Then the new `as_ptr ` has to be commented as well.

rikosellic:

rikosellic (approved):

PR #739, opened 2026-09-03
Prove simple kvirt area obligations

DID-Lab-SZU:


ostd/src/mm/kspace/kvirt_area.rs
```diff
@@ -367,7 +367,10 @@ impl KVirtArea {
         self.range.end
     }
 
-    pub fn range(&self) -> Range<Vaddr> {
+    pub fn range(&self) -> Range<Vaddr>
+        returns
```
Marsman1996: Does guideline say that prefer to use `verus_spec`?

Marsman1996 (changes requested):

ostd/src/mm/kspace/kvirt_area.rs
Marsman1996: And isn't this function already be proven

rikosellic:
LGTM, please merge the latest main.

rikosellic:

Marsman1996 (approved):
LGTM!

PR #740, opened 2026-09-03
disable unused sync examples

rikosellic:
Files in `spec/sync` are all early explorations of concurrency verification. This PR moves them into an `example` directory and disables its verification.


hiroki-chen:
Can you try rebase this PR from #737

rikosellic:
There's a rlimit error with upstream Verus. Anyway, it seems to be an independent issue. I will merge this PR.

PR #742, opened 2026-09-04
prove: id-alloc

Marsman1996:
sync the id-alloc to latest 0.18.1 and prove it.


rikosellic:
There are a lot of seemingly unrelated changes. Is this already solved by #743, so we do not need these changes?

Marsman1996:
> There are a lot of seemingly unrelated changes. Is this already solved by #743, so we do not need these changes?

Only partly solved. 
Though I do not know why, the `id-alloc` added in this PR makes the verification of [`mm::frame::segment` fail in macOS](https://github.com/asterinas/vostd/actions/runs/33938309862/job/101230500673?pr=742), and [`specs::mm::embedding` fail in Linux](https://github.com/asterinas/vostd/actions/runs/33938309878/job/101230500597?pr=742) fail.

rikosellic:
I have walked through the bitvec crate and find that its trait bounds like BitStore are pretty subtle. The current method that simply assuming specification for general trait bound is unsound. Can you just assume specification for specializied versions that only use primitive interger types (u8 u32 usize and u64 for 64-bit arch)as associated types, minmizing the TCB？

Marsman1996:
> Can you just assume specification for specializied versions that only use primitive interger types (u8 u32 usize and u64 for 64-bit arch)as associated types, minmizing the TCB？

The TCB has been narrowed to u8, u32, usize, and u64 on 64-bit targets.

verified_libs/vstd_extra/src/external/bitvec.rs
```diff
@@ -0,0 +1,338 @@
...
+//! below expose that model; every executed `bitvec` operation `id-alloc` performs is
+//! equated to a `Seq` operation here, so all reasoning in `id-alloc` stays at the
+//! `Seq<bool>` level for the admitted storage and order instances.
+//!
+//! The index and `get` operations are specified with
+//! `external_fn_specification` wrappers rather than `assume_specification`, because
+//! `BitVec`/`BitSlice` (foreign types) deref/index through associated trait types
+//! (`Deref::Target`, `Index::Output`, `BitSliceIndex::Immut`) whose reduction is
+//! matched through the original generic signature. Their contracts additionally
+//! require an admitted index type (`usize` for indexing, `Range<usize>` for `get`).
```
rikosellic: Do we need such a long explanation?

verified_libs/vstd_extra/src/external/bitvec.rs
```diff
@@ -0,0 +1,338 @@
...
+    axiom_usize_bitvec_index_model,
+    axiom_range_bitslice_get_model,
+    #[cfg(target_pointer_width = "64")]
+    axiom_u64_bitvec_model,
+}
+
+/// A `BitVec` derefs to a `BitSlice` over exactly its own bits. Specified with
+/// `assume_specification` (per-impl) so it overrides vstd's generic `Deref` trait
+/// spec, which otherwise loses the `bitslice_view` connection at auto-deref call
+/// sites (`bv.set(..)`, `bv.get(..)`).
```
rikosellic: This comment is also long, it is even unnecessary.

verified_libs/vstd_extra/src/external/bitvec.rs
```diff
@@ -0,0 +1,338 @@
...
+            &&& bitvec_view(final(bv)) == bitslice_view(final(ret))
+        },
+;
+
+/// Constructs an empty `BitVec` (length 0). The capacity hint is not modelled.
+pub assume_specification<T: BitStore, O: BitOrder>[ BitVec::<T, O>::with_capacity ](
+    capacity: usize,
+) -> (ret: BitVec<T, O>)
+    requires
+        obeys_bitvec_model::<T, O>(),
```
rikosellic, line 160: This function has a panic condition; it is not modeled here. You should add a precondition that ensures it will not panic (not the may_panic way). https://docs.rs/bitvec/latest/bitvec/vec/struct.BitVec.html#method.with_capacity

verified_libs/vstd_extra/src/external/bitvec.rs
```diff
@@ -0,0 +1,338 @@
...
+
+/// The number of bits.
+pub assume_specification<T: BitStore, O: BitOrder>[ BitVec::<T, O>::len ](
+    bv: &BitVec<T, O>,
+) -> (ret: usize)
+    requires
+        obeys_bitvec_model::<T, O>(),
+    ensures
+        ret == bitvec_view(bv).len(),
+;
```
rikosellic, line 196: Use `returns` instead of `ensures`.

verified_libs/vstd_extra/src/external/bitvec.rs
```diff
@@ -0,0 +1,338 @@
...
+/// Reads a single bit. Panics if `idx` is out of bounds. The postcondition is
+/// expressed abstractly via [`bitvec_index_value`]; the `usize` instance is related
+/// to the model by [`axiom_bitvec_index_usize`].
+pub uninterp spec fn bitvec_index_value<'a, T: BitStore, O: BitOrder, Idx>(
+    bv: &'a BitVec<T, O>,
+    idx: Idx,
+) -> &'a <BitVec<T, O> as Index<Idx>>::Output where BitSlice<T, O>: Index<Idx>;
+
+#[verifier(external_fn_specification)]
+pub fn bitvec_index<'a, T: BitStore, O: BitOrder, Idx>(bv: &'a BitVec<T, O>, idx: Idx) -> (ret:
```
rikosellic: Although explained, I'm still not clear why `assume_specification` does not work.

rikosellic (changes requested):

verified_libs/vstd_extra/src/external/bitvec.rs
Marsman1996: I guess they are added due to the `document-real-proof-debt` coding guide line

verified_libs/vstd_extra/src/external/bitvec.rs
Marsman1996: GPT-5.6-Sol < GLM 5.2
Today I'm using the GLM 5.2 and it successes with `assume_specification`...

verified_libs/vstd_extra/src/external/bitvec.rs
```diff
@@ -0,0 +1,317 @@
...
+;
+
+/// The number of bits.
+pub assume_specification<T: BitStore, O: BitOrder>[ BitVec::<T, O>::len ](
+    bv: &BitVec<T, O>,
+) -> (ret: usize)
+    requires
+        obeys_bitvec_model::<T, O>(),
+    returns
+        bitvec_view(bv).len() as usize,
```
rikosellic, line 195: Is `as usize` necessary? Also, the `ret` in `(ret: usize)` is redundant.

verified_libs/vstd_extra/src/external/bitvec.rs
```diff
@@ -0,0 +1,317 @@
...
+        bitvec_view(&ret).len() == 0,
+;
+
+/// Resizes the `BitVec` to `new_len`, filling new positions with `value` and
+/// preserving existing bits up to the shorter length.
+pub assume_specification<T: BitStore, O: BitOrder>[ BitVec::<T, O>::resize ](
+    bv: &mut BitVec<T, O>,
+    new_len: usize,
+    value: bool,
+) -> (ret: ())
```
rikosellic: Do we even need a returnn value here?

verified_libs/vstd_extra/src/external/bitvec.rs
Marsman1996: We need `as usize` here, otherwise there will be error
```
188 |           bitvec_view(bv).len(),
    |           ^^^^^^^^^^^^^^^^^^^^^ the `returns` clause has type `nat`
```

verified_libs/vstd_extra/src/external/bitvec.rs
```diff
@@ -0,0 +1,317 @@
...
+
+/// The content of a borrowed bit-slice, modelled as a sequence of booleans.
+pub uninterp spec fn bitslice_view<T: BitStore, O: BitOrder>(b: &BitSlice<T, O>) -> Seq<bool>;
+
+/// Whether storage and order support the immutable sequence model. Only the
+/// concrete instances in `group_bitvec_models` are trusted below.
+pub uninterp spec fn obeys_bitvec_model<T: BitStore, O: BitOrder>() -> bool;
+
+/// The only index and `get` specializations covered by this bridge.
+pub uninterp spec fn obeys_bitvec_index_model<Idx>() -> bool;
```
rikosellic: After more investigation, this should be `obeys_bitslice_index_model`. By the way, should this function signature be `<T,O,Idx>`
```
where
    T: BitStore,
    O:BitOrder.
   BitSlice<T,O>: Index<Idx>
```

matching the [documentation](https://docs.rs/bitvec/latest/bitvec/vec/struct.BitVec.html#impl-Index%3CIdx%3E-for-BitVec%3CT,+O%3E)?

verified_libs/vstd_extra/src/external/bitvec.rs
```diff
@@ -0,0 +1,317 @@
...
+    bv: &BitVec<T, O>,
+    idx: usize,
+)
+    requires
+        obeys_bitvec_model::<T, O>(),
+        idx < bitvec_view(bv).len(),
+    ensures
+        #![trigger bitvec_index_value(bv, idx)]
+        *bitvec_index_value(bv, idx) == bitvec_view(bv)[idx as int],
+;
```
rikosellic, line 224: You can directly write:
```
    requires
        obeys_bitvec_model::<T, O>(),
    ensures
        *bitvec_index_value(bv, idx) ==  bitvec_view(bv)[idx as int],
```
[Seq::spec_index](https://verus-lang.github.io/verus/verusdoc/vstd/seq/struct.Seq.html#method.spec_index) will simply return an arbitrary value when it is out of bounds. I think KVerus is not good enough to distinguish between spec and exec modes and fails to leverage existing vstd knowledge here.

verified_libs/vstd_extra/src/external/bitvec.rs
```diff
@@ -0,0 +1,317 @@
...
+/// A `BitVec`'s length is a `usize`, hence at most `usize::MAX`. The `len`
+/// `assume_specification` only fires at exec `.len()` calls, so without this axiom
+/// `bitvec_view(bv).len()` is unconstrained away from call sites (breaking overflow
+/// checks like `start + offset`).
+pub broadcast axiom fn axiom_bitvec_len_bound<T: BitStore, O: BitOrder>(bv: &BitVec<T, O>)
+    requires
+        obeys_bitvec_model::<T, O>(),
+    ensures
+        #![trigger bitvec_view(bv)]
+        bitvec_view(bv).len() <= usize::MAX as int,
```
rikosellic: This postcondition can be stronger, it is `usize::MAX/8` according to `with_capacity`.

verified_libs/vstd_extra/src/external/bitvec.rs
rikosellic: Ah, this is sensible.

verified_libs/vstd_extra/src/external/bitvec.rs
Marsman1996: have to keep `#![trigger bitvec_index_value(bv, idx)]` otherwise there will be a trigger warning

verified_libs/vstd_extra/src/external/bitvec.rs
```diff
@@ -0,0 +1,326 @@
...
+
+pub assume_specification<'a, T: BitStore, O: BitOrder, I: BitSliceIndex<'a, T, O>>[ BitSlice::<
+    T,
+    O,
+>::get ](bv: &'a BitSlice<T, O>, idx: I) -> (ret: Option<<I as BitSliceIndex<'a, T, O>>::Immut>)
+    requires
+        obeys_bitvec_model::<T, O>(),
+        obeys_bitslice_get_model::<'a, T, O, I>(),
+    ensures
+        ret == bitslice_get_value(bv, idx),
```
rikosellic: Use returns here.

verified_libs/vstd_extra/src/external/bitvec.rs
```diff
@@ -0,0 +1,326 @@
...
+    ensures
+        #![trigger bitslice_get_value(bv, range)]
+        match bitslice_get_value(bv, range) {
+            Some(s) => bitslice_view(s) == bitslice_view(bv).subrange(
+                range.start as int,
+                range.end as int,
+            ),
+            None => false,
+        },
+;
```
rikosellic, line 297: As in the `index` example, we don't need an `Option` here. A better way may be to delete the precondition and write the postcondition like
```
bitslice_get_value(..) = if 0 <= range.start <= range.end <= bitslice_view(bv).len() 
{bitslice_view.subrange(...)}
else
{arbitrary()}
```
The current code is more like an executable style.

verified_libs/vstd_extra/src/external/bitvec.rs
```diff
@@ -0,0 +1,326 @@
...
+            Some(j) => {
+                &&& 0 <= j < bitslice_view(bv).len()
+                &&& !bitslice_view(bv)[j as int]
+                &&& forall|i: int|
+                    #![trigger bitslice_view(bv)[i]]
+                    0 <= i < j ==> bitslice_view(bv)[i]
+            },
+            None => forall|i: int|
+                #![trigger bitslice_view(bv)[i]]
+                0 <= i < bitslice_view(bv).len() ==> bitslice_view(bv)[i],
```
rikosellic, line 313: I prefer `Seq::all()` here.

verified_libs/vstd_extra/src/external/bitvec.rs
```diff
@@ -0,0 +1,326 @@
...
+    requires
+        obeys_bitvec_model::<T, O>(),
+    ensures
+        match ret {
+            Some(j) => {
+                &&& 0 <= j < bitslice_view(bv).len()
+                &&& !bitslice_view(bv)[j as int]
+                &&& forall|i: int|
+                    #![trigger bitslice_view(bv)[i]]
+                    0 <= i < j ==> bitslice_view(bv)[i]
```
rikosellic: Not sure whether `bitslice_view(..).subrange(..).all(..)` will be nicer.

verified_libs/vstd_extra/src/external/bitvec.rs
Marsman1996: We need `Option` here, since `get` returns `Option<&BitSlice>`

verified_libs/vstd_extra/src/external/bitvec.rs
Marsman1996: GLM 5.2 does not think it will be nicer 🤔

verified_libs/vstd_extra/src/external/bitvec.rs
Marsman1996: Use `Seq::all()` and introduce a new `axiom fn axiom_bitslice_all_true`

verified_libs/vstd_extra/src/external/bitvec.rs
rikosellic: Oh, it is sensible.

verified_libs/vstd_extra/src/external/bitvec.rs
```diff
@@ -0,0 +1,341 @@
...
+    ensures
+        #[trigger] obeys_bitvec_model::<usize, Lsb0>(),
+;
+
+pub broadcast axiom fn axiom_usize_bitslice_index_model<T: BitStore, O: BitOrder>()
+    requires
+        obeys_bitvec_model::<T, O>(),
+    ensures
+        #[trigger] obeys_bitslice_index_model::<T, O, usize>(),
+;
```
rikosellic, line 114: Can we add a macro to generate these specialized axioms to save some lines?

verified_libs/vstd_extra/src/external/bitvec.rs
```diff
@@ -0,0 +1,341 @@
...
+        obeys_bitvec_model::<T, O>(),
+    ensures
+        #![trigger <BitVec<T, O> as IndexSpec<usize>>::index_req(bv, &i)]
+        <BitVec<T, O> as IndexSpec<usize>>::index_req(bv, &i) == (i < bitvec_view(bv).len()),
+;
+
+/// Bit length is bounded by `BitSlice::<T, O>::MAX_BITS` (= `usize::MAX >> 3`). The
+/// `len` `assume_specification` only fires at exec `.len()` calls, so this axiom
+/// constrains `bitvec_view(bv).len()` away from call sites (e.g. for overflow checks).
+pub broadcast axiom fn axiom_bitvec_len_bound<T: BitStore, O: BitOrder>(bv: &BitVec<T, O>)
```
rikosellic, line 237: The comment is a bit wordy. The first line is informative enough.

verified_libs/vstd_extra/src/external/bitvec.rs
rikosellic: Thanks for the correction. But I still think we can rephrase the postcondition like this
```
bitslice_get_value(bv, range) == 
    if 0 <= range.start <= range.end <= bitslice_view(bv).len() 
    { Some(subrange) }
    else
    None
```
This way interprets the definition more cleanly.

verified_libs/vstd_extra/src/external/bitvec.rs
```diff
@@ -0,0 +1,341 @@
...
+/// `broadcast proof fn` would self-reveal-cycle.
+pub broadcast axiom fn axiom_bitslice_all_true<T: BitStore, O: BitOrder>(bv: &BitSlice<T, O>)
+    requires
+        obeys_bitvec_model::<T, O>(),
+    ensures
+        #![trigger bitslice_view(bv)]
+        bitslice_view(bv).all(|b| b) ==> forall|i: int|
+            #![trigger bitslice_view(bv)[i]]
+            0 <= i < bitslice_view(bv).len() ==> bitslice_view(bv)[i],
+;
```
rikosellic: What's happening here. First, this statement seems to be provable.  Also, why do we need this lemma, is it caused by a trigger problem? If so, maybe we have to use the explicit `forall` approach directly.

rikosellic (changes requested):

verified_libs/vstd_extra/src/external/bitvec.rs
Marsman1996: `&BitSlice` is opaque, so the `Some(subrange)` value can't be constructed — the literal `== if {Some(...)} else {None}` form isn't writable, and the per-arm match is the faithful equivalent.

verified_libs/vstd_extra/src/external/bitvec.rs
Marsman1996: due to the lifetime diff between `axiom_usize_bitslice_index_model` and `axiom_range_bitslice_get_model`, it will not save lines, however, I add macro for `axiom_u8/u32/usize_bitvec_model`

ostd/specs/mm/embedding/mod.rs
```diff
@@ -3882,6 +3882,7 @@ proof fn lemma_step_segment_clone_range<'rcu>(
     assert(s.regions.inv()) by {
         reveal(VmStore::structural_inv);
     };
+    let ghost old_store = *s;
```
rikosellic: Is this definition used?

Marsman1996: I think this is introduced by the merge...
And not used anymore

ostd/libs/id-alloc/src/lib.rs
```diff
@@ -1,65 +1,349 @@
 // SPDX-License-Identifier: MPL-2.0
-
 #![cfg_attr(not(test), no_std)]
 #![deny(unsafe_code)]
+#![feature(proc_macro_hygiene)]
+#![expect(internal_features)]
+
+use ::bitvec::prelude::BitVec;
```
rikosellic: Is the `::` at the beginning necessary? Besides, this import is under `core::{fmt::Debug, ops::Range}` in the source code.

ostd/libs/id-alloc/src/lib.rs
```diff
@@ -1,65 +1,349 @@
...
 
-use bitvec::prelude::BitVec;
+verus! {
+
+/// The index of the first `false` bit in `s`, or `s.len()` if every bit is `true`.
+pub open spec fn is_first_zero(s: Seq<bool>, i: int) -> bool {
+    &&& 0 <= i <= s.len()
+    &&& (forall|j: int| #![trigger s.index(j)] 0 <= j < i ==> s.index(j))
+    &&& (i < s.len() ==> !s.index(i))
+}
```
rikosellic: Should we move this function into `external/bitvec` to save some lines in the postcondition of `first_zero`?

rikosellic: Should we move this function into `external/bitvec` to save some lines in the postcondition of `first_zero`? Besides, we can use `s[i]` and `s[j]` instead of `.index`

ostd/libs/id-alloc/src/lib.rs
```diff
@@ -1,65 +1,349 @@
...
+    decreases s.len(),
+{
+    if s.len() == 0 {
+        0
+    } else if !s.index(0) {
+        0
+    } else {
+        1 + first_zero_seq(s.subrange(1, s.len() as int))
+    }
+}
```
rikosellic: This one can also be moved into `external/bitvec`.

ostd/libs/id-alloc/src/lib.rs
```diff
@@ -1,65 +1,349 @@
...
+    type V = Seq<bool>;
+
+    closed spec fn view(&self) -> Seq<bool> {
+        bitvec_view(&self.bitset)
+    }
+}
+
+impl Inv for IdAlloc {
+    /// The well-formedness invariant: `first_available_id` is the first free bit.
+    closed spec fn inv(self) -> bool {
```
rikosellic, line 54: Can we define this as a `type_invariant`? It is okay if not, I can imagine it is temporarily broken during internal execution.

ostd/libs/id-alloc/src/lib.rs
```diff
@@ -1,65 +1,349 @@
...
+/// The index of the first `false` bit in `s`, or `s.len()` if every bit is `true`.
+pub open spec fn is_first_zero(s: Seq<bool>, i: int) -> bool {
+    &&& 0 <= i <= s.len()
+    &&& (forall|j: int| #![trigger s.index(j)] 0 <= j < i ==> s.index(j))
+    &&& (i < s.len() ==> !s.index(i))
+}
+
+/// Index of the first `false` bit, or `s.len()` if every bit is `true`. Defined
+/// recursively so it is deterministic and the SMT solver can unfold it.
+pub open spec fn first_zero_seq(s: Seq<bool>) -> int
```
rikosellic: This function name sounds a bit weird, but I have not come up with a better name.

ostd/libs/id-alloc/src/lib.rs
rikosellic, line 1: All the proofs in this file is using `,index` instead of `s[i]`.

ostd/libs/id-alloc/src/lib.rs
```diff
@@ -1,65 +1,349 @@
...
+    ensures
+        first_zero_seq(s.update(k - 1, true)) == k + first_zero_seq(s.subrange(k, s.len() as int)),
+{
+    let t = s.update(k - 1, true);
+    lemma_first_zero_seq_prefix_all_true(t, k);
+    assert(t.subrange(k, s.len() as int) =~= s.subrange(k, s.len() as int)) by {}
+}
+
+/// Element-wise-equal sequences have the same `first_zero_seq`.
+proof fn lemma_first_zero_seq_ext_equal(a: Seq<bool>, b: Seq<bool>)
```
rikosellic: Do we need this lemma? It seems very trivial.

rikosellic (changes requested):

ostd/libs/id-alloc/src/lib.rs
Marsman1996: No we can't. As you said it's temporarily broken

ostd/libs/id-alloc/src/lib.rs
Marsman1996: This is because `pub mod bitvec;` in verified_libs/vstd_extra/src/external/mod.rs, I have modified it to remove the `pub` token

ostd/libs/id-alloc/src/lib.rs
Marsman1996: removed

ostd/libs/id-alloc/src/lib.rs
```diff
@@ -1,65 +1,311 @@
...
+{
+    if i < j {
+        assert(false);
+    } else if j < i {
+        assert(false);
+    }
+}
+
+/// `first_zero_index(s)` itself satisfies `is_first_zero` (induction on `s.len()`).
+proof fn lemma_first_zero_index_char(s: Seq<bool>)
```
rikosellic: This name is not informative, maybe `lemma_first_zero_index_is_first_zero`

ostd/libs/id-alloc/src/lib.rs
```diff
@@ -1,65 +1,311 @@
...
-    ///
-    /// The slice of `u8` bytes is the raw data of a bitmap.
-    pub fn from_bytes_with_capacity(slice: &[u8], capacity: usize) -> Self {
-        let bitset = if capacity > slice.len() * 8 {
-            let mut bitset = BitVec::from_slice(slice);
-            bitset.resize(capacity, false);
-            bitset
+/// If the prefix `[0, k)` of `s` is all `true`, then the first zero of `s` is
+/// `k` plus the first zero of the remainder (induction on `s.len()`).
+proof fn lemma_first_zero_index_prefix_all_true(s: Seq<bool>, k: int)
```
rikosellic: The name of the lemma seems irrelevant with what it proves.

ostd/libs/id-alloc/src/lib.rs
```diff
@@ -73,41 +319,94 @@ impl IdAlloc {
...
+                &&& final(self).inv()
+            },
+    )]
     pub fn alloc_consecutive(&mut self, count: usize) -> Option<Range<usize>> {
         if count == 0 {
             return None;
         }
 
-        // Scan the bitmap from the position `first_available_id`
-        // for the first `count` number of consecutive 0's.
```
rikosellic, line 82: Why the comment is deleted?

ostd/libs/id-alloc/src/lib.rs
```diff
@@ -73,41 +319,94 @@ impl IdAlloc {
...
+        let end = self.first_available_id.checked_add(count)?;
+        if end > self.bitset.len() {
+            return None;
+        }
+
         let allocated_range = {
-            // Invariance: all bits within `curr_range` are 0's
```
rikosellic, line 84: This comment is also removed

ostd/libs/id-alloc/src/lib.rs
```diff
@@ -73,41 +319,94 @@ impl IdAlloc {
...
+                curr_range.start <= curr_range.end,
+                curr_range.end <= self@.len(),
+                range_usize_len_spec(&curr_range) <= count,
+                forall|j: int| #![trigger self@[j]] curr_range.start as int <= j < curr_range.end as int ==> !self@[j],
+                decreases self@.len() as int - curr_range.end as int,
+            )]
+            /* `Range::len` is unspecced by vstd.
+             * Origin Rust: while curr_range.len() < count && curr_range.end < self.bitset.len() {
+             */
+            while range_usize_len(&curr_range) < count && curr_range.end < self.bitset.len() {
```
rikosellic: How about adding `assume_specification` for it? Actually, it would be good to propose a PR to Verus.

ostd/libs/id-alloc/src/lib.rs
```diff
@@ -73,41 +319,94 @@ impl IdAlloc {
...
+            0 <= allocated_range.start,
+            allocated_range.end <= self@.len(),
+            allocated_range.start <= id,
+            id <= allocated_range.end,
+            forall|j: int| #![trigger self@[j]] allocated_range.start as int <= j < id as int ==> self@[j],
+            forall|j: int| 0 <= j < self@.len() && !(allocated_range.start as int <= j < id as int) ==> self@[j] == old(self)@[j],
+        )]
+        /* `Range::clone` is unspecced; iterate `start..end` directly.
+         * Origin Rust: for id in allocated_range.clone()
+         */
```
rikosellic: Can we assume_specfication for `Range::clone`? We can also create a PR.

rikosellic (changes requested):

ostd/libs/id-alloc/src/lib.rs
Marsman1996: We already have a spec in `ostd/src/mm/page_table/cursor/locking.rs` by Sean, I'm reusing it

ostd/libs/id-alloc/src/lib.rs
Marsman1996: It seems we cannot add `assume_specification` for it. Since it is a trait method
```
error: The verifier does not yet support the following Rust feature: assume_specification for a provided trait method
```

rikosellic: Indeed, that is a bit surprising. Looks sensible.

verified_libs/vstd_extra/src/external/bitvec.rs
```diff
@@ -0,0 +1,337 @@
...
+
+/// The index of the first `false` bit in `s`, or `s.len()` if every bit is `true`.
+pub open spec fn is_first_zero(s: Seq<bool>, i: int) -> bool {
+    &&& 0 <= i <= s.len()
+    &&& (forall|j: int| #![trigger s[j]] 0 <= j < i ==> s[j])
+    &&& (i < s.len() ==> !s[i])
+}
+
+/// Index of the first `false` bit, or `s.len()` if every bit is `true`. Defined
```
rikosellic: I suddenly realize the correct place for these `spec fn`s is  `seq_extra`.

ostd/libs/id-alloc/src/lib.rs
```diff
@@ -1,65 +1,311 @@
...
+    /// The well-formedness invariant: `first_available_id` is the first free bit.
+    closed spec fn inv(self) -> bool {
+        &&& 0 <= self.first_available_id <= self@.len()
+        &&& self.first_available_id == first_zero_index(self@)
+    }
+}
+
+/// `is_first_zero` is uniquely satisfied. Broadcast so it auto-fires whenever two
+/// `is_first_zero` facts arise.
+pub broadcast proof fn lemma_is_first_zero_unique(s: Seq<bool>, i: int, j: int)
```
rikosellic: The lemmas in this file can also be moved to `seq_extra`.

ostd/libs/id-alloc/src/lib.rs
Marsman1996: to `lemma_first_zero_index_after_true_prefix`

ostd/libs/id-alloc/src/lib.rs
```diff
@@ -1,65 +1,311 @@
...
+    requires
+        is_first_zero(s, i),
+        is_first_zero(s, j),
+    ensures
+        #![auto]
+        i == j,
+{
+    if i < j {
```
rikosellic: Do this function need any proofs? The proof code seems trivial. I'm also wondering whether we need to mark it as `broadcast`.

Marsman1996: weird, I will run simplify again

verified_libs/vstd_extra/src/external/range.rs
```diff
@@ -45,6 +45,12 @@ pub fn range_usize_is_empty(r: &Range<usize>) -> (ret: bool)
     !(r.start < r.end)
 }
 
+/// `Range::clone` copies the range; the result equals the original value.
+pub assume_specification<Idx: Clone>[ Range::<Idx>::clone ](range: &Range<Idx>) -> (res: Range<Idx>)
+    ensures
+        res == *range,
```
rikosellic: Unfortunately, I believe this is not correct. One trial may be
```
pub assume_specification<Idx: Clone>[ Range::<Idx>::clone ](range: &Range<Idx>) -> (res: Range<Idx>)
    ensures
        Idx::clone.ensures((&range.start,), res.start),
        Idx::clone.ensures((&range.end,), res.end),;
```

Marsman1996: Then we are using this wrong spec for a long time 👀

rikosellic: A better one maybe
```
pub assume_specification<Idx: Clone>[ Range::<Idx>::clone ](range: &Range<Idx>) -> (res: Range<Idx>)
    ensures
        Idx::clone.requires((&range.start,)) && Idx::clone.requires((&range.end,)) ==>
        {
            &&& Idx::clone.ensures((&range.start,), res.start)
            &&& Idx::clone.ensures((&range.end,), res.end)
        };
```

rikosellic: > Then we are using this wrong spec for a long time 👀

Indeed. The wrong spec is correct for primitive types we used in `ostd`, but it does not hold in general.

verified_libs/vstd_extra/src/seq_extra.rs
```diff
@@ -233,3 +233,222 @@ pub broadcast group group_seq_extra_lemmas {
...
+}
+
+/// Setting the bit at `k - 1` (the current first zero) to `true`, when the prefix
+/// `[0, k - 1)` is all `true`, advances the first zero to
+/// `k + first_zero_index(s.subrange(k, len))`.
+pub proof fn lemma_first_zero_index_advance_after_set(s: Seq<bool>, k: int)
+    requires
+        0 < k <= s.len(),
+        forall|j: int| 0 <= j < k - 1 ==> s[j],
+        k - 1 < s.len() ==> !s[k - 1],
```
rikosellic: This precondition is exactly `is_first_zero`

Marsman1996: 😣

dv
rikosellic, line 1: Is this changed dv update to the latest one？

Marsman1996: Yes, `a8834cbb0dc227e83d05033a452dc95875d94fd5` is the latest one, the original `dv` cannot deal with the `-` in package

ostd/libs/id-alloc/src/lib.rs
```diff
@@ -73,41 +118,96 @@ impl IdAlloc {
     /// If allocation is not possible, it returns `None`.
     ///
     /// TODO: Choose a more efficient strategy.
+    #[verus_spec(res =>
+        requires
+            old(self).inv(),
+        ensures
+            res matches Some(r) ==> {
+                &&& r.end - r.start == count
+                &&& 0 <= r.start
```
rikosellic: Since `r.start` is `usize`, I don't think this is necessary.

ostd/libs/id-alloc/src/lib.rs
```diff
@@ -118,53 +218,118 @@ impl IdAlloc {
...
+            forall|i: int| #![trigger final(self)@[i]] range.start <= i < range.end ==> !final(self)@[i],
+            forall|i: int|
+                0 <= i < final(self)@.len() && !(range.start <= i < range.end) ==> final(self)@[i] == old(self)@[i],
+            final(self).inv(),
+    )]
     pub fn free_consecutive(&mut self, range: Range<usize>) {
-        if range.is_empty() {
+        /* `Range::is_empty` is unspecced by vstd.
+         * Origin Rust: if range.is_empty() {
+         */
```
rikosellic: `Range::is_empty` is specified in #718, can we define the same assumption here? #718 should be rebased when this PR is merged into main.

ostd/libs/id-alloc/src/lib.rs
```diff
@@ -118,53 +218,118 @@ impl IdAlloc {
...
     ///
     /// If the `id` is out of bounds, this method will panic.
+    #[verus_spec(res =>
+        requires
+            old(self).inv(),
+            id < self@.len(),
+        ensures
+            res is Some ==> {
+                &&& final(self)@ == old(self)@.update(id as int, true)
+                &&& res == (if old(self)@[id as int] { None } else { Some(id) })
```
rikosellic: This postcondition should be distributed to the two branches. The current definition is not strong enough.

ostd/libs/id-alloc/src/lib.rs
```diff
@@ -174,13 +339,45 @@ impl IdAlloc {
     /// # Panics
     ///
     /// If the `id` is out of bounds, this method will panic.
+    #[verus_spec(ret =>
+        requires
+            id < self@.len(),
+        ensures
```
rikosellic: Use `returns` here.

rikosellic:
#753 has greatly improved the performance of `lemma_step_segment_clone_range` in `embedding/mod.rs`. Please wait for that PR and see whether the modification here is still necessary.

Marsman1996:
let me resolve this conflict

Marsman1996:
Need to add `requires` in util::ops since the spec of `range::clone` updated

ostd/libs/id-alloc/src/lib.rs
rikosellic: Update: vstd already supports `Range::len` now; no need to add any assumptions. I will remove this function in #756

Marsman1996:
The IRC11 commit need to be updated🤔

rikosellic (approved):

PR #744, opened 2026-09-04
ci: add verus update test under macOS

Marsman1996:
To prevent the Verus which is failing the macOS to be merged (see https://github.com/asterinas/vostd/pull/743)


rikosellic:

PR #748, opened 2026-09-08
chore: minor improvement in spec

rikosellic:


rikosellic:
There seems to be a significant verification performance regression.

PR #751, opened 2026-09-09
simplify: mm::page_table::cursor

Marsman1996:
simplify ~1000 LoC


rikosellic:
It's like losing weight. I love it.

PR #755, opened 2026-09-10
simplify: mm::page_table::cursor

Marsman1996:
Once again KVerus simplified ~800 LoC, hopefully this time will not slow the verification down


Marsman1996:
Man, what can I say

PR #766, opened 2026-09-14
ci: refactor ci workflow

Marsman1996:
See https://github.com/Marsman1996/vostd/actions/runs/34851126760 for the example result, now all in one, and reduce 2 runs


rikosellic:
Very elegant!

PR #770, opened 2026-09-16
prove: cpu::set

Marsman1996:
The exec code need the modification in Verus for the spec of `iter:: Enumerate`, `iter:: FilterMap`, `iter::FlatMap` and `iter::sum`.

Potential useful PR:
`iter:: Enumerate`: https://github.com/verus-lang/verus/pull/2904
`iter:: FilterMap`: https://github.com/verus-lang/verus/pull/2728
`iter::sum`: https://github.com/verus-lang/verus/pull/2962


verified_libs/vstd_extra/src/external/int_specs.rs
```diff
@@ -41,4 +41,15 @@ pub assume_specification[ usize::is_power_of_two ](self_: usize) -> (r: bool)
     no_unwind
 ;
 
+/// `usize::div_ceil` panics if `rhs` is zero. The precondition excludes that case,
+/// so every valid call satisfies `no_unwind`.
+pub assume_specification[ usize::div_ceil ](self_: usize, rhs: usize) -> (r: usize)
+    requires
+        rhs > 0,
+    ensures
+        (r as int) == ((self_ as int) + (rhs as int) - 1) / (rhs as int),
```
rikosellic: I believe these `as int` are redundant.

verified_libs/vstd_extra/src/external/int_specs.rs
```diff
@@ -41,4 +41,15 @@ pub assume_specification[ usize::is_power_of_two ](self_: usize) -> (r: bool)
...
+    opens_invariants none
+    no_unwind
```
rikosellic: We'd better remove this `no_unwind` , this function can indeed panic, although not when the precondition is satisfied.

verified_libs/vstd_extra/src/external/smallvec.rs
```diff
@@ -0,0 +1,174 @@
...
+;
+
+/// Constructs a new, empty `SmallVec` with the given heap capacity, which is not modelled.
+///
+/// Panics on capacity overflow if `n * size_of::<A::Item>()` exceeds `isize::MAX`
+/// (via `reserve_exact`, which requests exactly `n` elements).
+pub assume_specification<A: Array>[ SmallVec::<A>::with_capacity ](n: usize) -> (ret: SmallVec<A>)
+    requires
+        obeys_smallvec_array::<A>(),
+        (n as int) * (size_of::<A::Item>() as int) <= isize::MAX as int,
```
rikosellic: First, `as int` seems unnecessary. Second, according to the source code, it does not seem to constrain the `usize`? The comment seems reasonable, but I do not find any `isize::MAX` constraint in the internal `reseve_exact` or the further inner `try_grow`.

verified_libs/vstd_extra/src/external/smallvec.rs
```diff
@@ -0,0 +1,174 @@
...
+    ensures
+        smallvec_view(&ret) == Seq::<A::Item>::empty(),
+;
+
+/// The number of elements.
+pub assume_specification<A: Array>[ SmallVec::<A>::len ](v: &SmallVec<A>) -> (len: usize)
+    requires
+        obeys_smallvec_array::<A>(),
+    ensures
+        (len as int) == smallvec_view(v).len(),
```
rikosellic: `as int`. We may need to add this in our guidline, or KVerus skills?

rikosellic (changes requested):

verified_libs/vstd_extra/src/external/smallvec.rs
Marsman1996: later in the guideline I think

verified_libs/vstd_extra/src/external/int_specs.rs
Marsman1996: how about `no_unwind when rhs > 0`?

rikosellic: That sounds good! I forgot about this feature.

verified_libs/vstd_extra/src/external/smallvec.rs
Marsman1996: with_capacity → reserve_exact → try_reserve_exact → try_grow → layout_array → Layout::from_size_align

```rust
  fn layout_array<T>(n: usize) -> Result<Layout, CollectionAllocErr> {
      let size = mem::size_of::<T>()
          .checked_mul(n)
          .ok_or(CollectionAllocErr::CapacityOverflow)?;        // usize overflow
      let align = mem::align_of::<T>();
      Layout::from_size_align(size, align)                       // <-- isize::MAX size-overflow
          .map_err(|_| CollectionAllocErr::CapacityOverflow)
  }
```

and because it is infallible, then it will panic,
see https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=ab623204886a12818e192c6a29993d52

rikosellic: That's very convincing! Better than the docimentation and a human reader, thanks!

verified_libs/vstd_extra/src/external/smallvec.rs
```diff
@@ -0,0 +1,173 @@
...
+        n * size_of::<A::Item>() <= isize::MAX,
+    ensures
+        smallvec_view(&ret) == Seq::<A::Item>::empty(),
+;
+
+/// The number of elements.
+pub assume_specification<A: Array>[ SmallVec::<A>::len ](v: &SmallVec<A>) -> (len: usize)
+    requires
+        obeys_smallvec_array::<A>(),
+    ensures
```
rikosellic, line 125: Returns is better.

rikosellic (changes requested):

verified_libs/vstd_extra/src/external/bits.rs
```diff
@@ -0,0 +1,189 @@
...
+
+/// Masking with the low-`k`-bits mask `(1u64 << k) - 1` clears bit `b >= k` of any
+/// word (AND associativity folded into the statement).
+///
+/// Kept as an axiom: the hypothesis embeds the subtraction `(1u64 << k) - 1`, which
+/// spec-mode arithmetic promotes to `int` (E0308 feeding `&`), so neither the mask
+/// value nor a subtraction-shaped hypothesis can enter a `bit_vector` formula. A
+/// bit-only rewiring of the whole family (`mask == !(!0u64 << k)`) would force
+/// rewriting the caller's mask terms and re-bridging the exec-computed mask value.
+pub broadcast axiom fn axiom_u64_masked_bit_clear(word: u64, mask: u64, k: int, b: int)
```
rikosellic: Can we get reid of these axioms?

verified_libs/vstd_extra/src/external/bits.rs
```diff
@@ -0,0 +1,189 @@
...
+    lemma_u64_setbit_bit_set,
+    lemma_u64_setbit_bit_keep,
+    lemma_u64_clearbit_bit_clear,
+    lemma_u64_clearbit_bit_keep,
+    lemma_u64_zero_and_bit,
+}
+
+/// The number of set bits in a `u64` word: the uninterpreted spec carrier for
+/// `count_ones` and the counting bridge.
+pub uninterp spec fn u64_set_bits(w: u64) -> int;
```
rikosellic: I'm not very happy about this. Can you define this function? Although it may be complicated.

ostd/src/cpu/mod.rs
```diff
@@ -97,7 +102,59 @@ pub fn all_cpus() -> impl Iterator<Item = CpuId> {
     (0..num_cpus()).map(|id| CpuId(id as u32))
 }
 
-cpu_local_cell! {
+verus! {
+
+/// The underlying numeric CPU id.
+impl View for CpuId {
+    type V = u32;
```
rikosellic: Will it be simpler to use `int`?

ostd/src/cpu/mod.rs
```diff
@@ -97,7 +102,59 @@ pub fn all_cpus() -> impl Iterator<Item = CpuId> {
...
+/// hence also a `usize`; the `u32` bound is what lets a `CpuSet` bitset fit in its backing
+/// `SmallVec` without overflowing `isize::MAX`.
+pub broadcast axiom fn axiom_cpu_count_bounds()
+    ensures
+        #![trigger cpu_count()]
+        1 <= cpu_count() <= u32::MAX as int,
+;
+
+/// The documented `CpuId` type invariant: the id is less than the CPU count.
+impl Inv for CpuId {
```
rikosellic: Can we just define a type invariant?

rikosellic (changes requested):

verified_libs/vstd_extra/src/external/bits.rs
Marsman1996: GPT-5.6 Sol can prove these, better than GLM-5.3-Flash (which is no surprise)

verified_libs/vstd_extra/src/external/bits.rs
rikosellic, line 1: The lemmas except those related to `u64_set_bit` can be moved to a new `vstd_extra/bits` module. BTW, many lemmas has a postcondition like `0 <= k < 64 ==>...`, can we move `0<=k < 64` to `requires`?

ostd/src/cpu/mod.rs
```diff
@@ -97,7 +101,54 @@ pub fn all_cpus() -> impl Iterator<Item = CpuId> {
...
+    #[verifier::type_invariant]
+    closed spec fn type_inv(self) -> bool {
+        0 <= self@ < cpu_count()
+    }
+}
+
+/// The number of CPUs; binds the exec `num_cpus()` to the trusted `cpu_count()`.
+pub assume_specification[ crate::cpu::num_cpus ]() -> (n: usize)
+    ensures
+        (n as int) == cpu_count(),
```
rikosellic: I believe we can use `returns` here.

ostd/src/cpu/mod.rs
```diff
@@ -97,7 +101,54 @@ pub fn all_cpus() -> impl Iterator<Item = CpuId> {
...
+/// is the explicit specification boundary (mirrors io_port's num_cpus-style pattern).
+pub uninterp spec fn cpu_count() -> int;
+
+/// The count is a positive value that fits a `u32` (the exec reads `static NUM_CPUS: u32`),
+/// hence also a `usize`; the `u32` bound is what lets a `CpuSet` bitset fit in its backing
+/// `SmallVec` without overflowing `isize::MAX`.
+pub broadcast axiom fn axiom_cpu_count_bounds()
+    ensures
+        #![trigger cpu_count()]
+        1 <= cpu_count() <= u32::MAX as int,
```
rikosellic: Unnecessary `as int`.

ostd/src/cpu/mod.rs
```diff
@@ -97,7 +101,54 @@ pub fn all_cpus() -> impl Iterator<Item = CpuId> {
...
+    closed spec fn view(&self) -> int {
+        self.0 as int
+    }
+}
+
+/// The trusted value of the system CPU count. The exec reads `NUM_CPUS`, a `u32`
+/// initialized to 1 and set once at boot (never decreased, no hot-plug), so it is a
+/// stable constant; Verus cannot mention an `exec static` in a spec, so this uninterp
+/// is the explicit specification boundary (mirrors io_port's num_cpus-style pattern).
+pub uninterp spec fn cpu_count() -> int;
```
rikosellic, line 125: This is a very similar situation with `max_paddr`. Maybe later we will have a `spec/config` module that holds all definitions like this. LGTM for now.

ostd/src/cpu/mod.rs
```diff
@@ -97,7 +101,54 @@ pub fn all_cpus() -> impl Iterator<Item = CpuId> {
...
+pub assume_specification[ crate::cpu::num_cpus ]() -> (n: usize)
+    ensures
+        (n as int) == cpu_count(),
+;
+
+/// `CpuId::as_usize`; its integer value is the abstract CPU id.
+pub assume_specification[ crate::cpu::CpuId::as_usize ](c: CpuId) -> usize
+    returns
+        c@ as usize,
+;
```
rikosellic: I'm a bit confused here, why do we need to `assume_specification` for such a simple function.

rikosellic (changes requested):

ostd/src/cpu/mod.rs
Marsman1996: no need indeed

verified_libs/vstd_extra/src/bits.rs
rikosellic, line 1: Are all the axioms defined here used?

ostd/src/cpu/mod.rs
```diff
@@ -97,7 +103,48 @@ pub fn all_cpus() -> impl Iterator<Item = CpuId> {
...
+impl CpuId {
+    /// A CPU id is always in the range of CPUs present in the system.
+    #[verifier::type_invariant]
+    closed spec fn type_inv(self) -> bool {
+        0 <= self@ < cpu_count()
+    }
+}
+
+/// The number of CPUs; binds the exec `num_cpus()` to the trusted `cpu_count()`.
+pub assume_specification[ crate::cpu::num_cpus ]() -> usize
```
rikosellic: Can we use `external_body` instead of `assume_specification`?

ostd/src/cpu/set.rs
```diff
@@ -9,143 +26,752 @@ use crate::const_assert;
...
 const fn part_idx(cpu_id: CpuId) -> usize {
+    proof! {
+        reveal(part_idx_spec);
+    }
     cpu_id.as_usize() / BITS_PER_PART
 }
 
+#[verus_verify]
+#[verus_spec(
+    returns bit_idx_spec(cpu_id) as usize,
```
rikosellic: Can we use `dual_spec` here?

rikosellic: Or `when_used_as_spec`.

ostd/src/cpu/set.rs
```diff
@@ -9,143 +26,752 @@ use crate::const_assert;
...
+    proof! {
+        reveal(bit_idx_spec);
+    }
     cpu_id.as_usize() % BITS_PER_PART
 }
 
+#[verus_verify]
+#[verus_spec(
+    returns parts_for_cpus_spec(num_cpus as int) as usize,
+)]
```
rikosellic, line 63: Like `part_idx`.

rikosellic (changes requested):

ostd/src/cpu/set.rs
Marsman1996: not `bit_idx`, since there is `as_usize`.
Using `when_used_as_spec`

ostd/src/cpu/set.rs
Marsman1996: same, because of `div_ceil`

verified_libs/vstd_extra/src/bits.rs
Marsman1996: Remove any of them will cause the proof error

verified_libs/vstd_extra/src/bits.rs
```diff
@@ -0,0 +1,155 @@
...
+        b2 as usize))),
+{
+    let bu: u32 = b as u32;
+    let b2u: u32 = b2 as u32;
+    assert((bu != b2u) ==> (((word & (!(1u64 << bu))) & (1u64 << b2u)) == (word & (1u64 << b2u))))
+        by (bit_vector);
+}
+
+/// AND-ing the zero word with any in-range unit bit is zero.
+pub broadcast proof fn lemma_u64_zero_and_bit(k: int)
```
rikosellic: Do we really need such a specific lemma? `0 & _` will be 0.

ostd/src/cpu/set.rs
```diff
@@ -9,143 +26,753 @@ use crate::const_assert;
...
+    lemma_u64_set_bits_nonzero,
+    group_u64_bit_algebra,
+    crate::cpu::axiom_cpu_count_bounds,
+    vstd::layout::layout_of_primitives,
+    vstd::set::group_set_lemmas,
+    vstd::set_lib::range_set_properties,
+};
+
+/// `div_ceil(n, BITS_PER_PART)`: the number of 64-bit words needed to hold `n` bits.
+pub closed spec fn parts_for_cpus_spec(n: int) -> int {
```
rikosellic: Why don't we just keep one `parts_for_cpus_spec` from `size` to `size`

ostd/src/cpu/set.rs
```diff
@@ -9,143 +26,753 @@ use crate::const_assert;
...
+            idx += 1;
+        }
+        count
     }
 
     /// Returns true if the set is empty.
+    #[verus_spec(ret =>
+        requires
+            self.inv(),
+        ensures
```
rikosellic: returns

ostd/src/cpu/set.rs
```diff
@@ -9,143 +26,753 @@ use crate::const_assert;
...
+        assert(parts_for_cpus_spec(cpu_count()) == (cpu_count() + 63) / 64) by {
+            reveal(parts_for_cpus_spec)
+        };
+    }
+}
+
+impl Inv for CpuSet {
+    /// The backing vector holds exactly `parts_for_cpus(cpu_count)` words, the view
+    /// is bounded by `cpu_count`, and the unused tail bits (>= `cpu_count`) are clear.
+    closed spec fn inv(self) -> bool {
```
rikosellic, line 132: Make this a type_invariant

ostd/src/cpu/set.rs
```diff
@@ -9,143 +26,753 @@ use crate::const_assert;
...
+        ((1u64 << ((num_cpus % 64) as usize)) - 1) as u64
+    } else {
+        !0u64
+    }
+}
+
+/// Bit `i` (`i % 64` of word `i / 64`) is set in the bit sequence `seq`.
+pub open spec fn bit_at(seq: Seq<u64>, i: int) -> bool {
+    if 0 <= i < 64 * seq.len() as int {
+        (seq[i / 64] & (1u64 << ((i % 64) as usize))) != 0
```
rikosellic: Why don't we reuse `part_idx` and `bit_idx` here? The current definition is rather confusing. We should do this in other spec fn too.

ostd/src/cpu/set.rs
rikosellic, line 1: Proofs in this file are very wordy. Should be simplified.

rikosellic (changes requested):

ostd/src/cpu/set.rs
Marsman1996: This file has been run strip skill, and most of them can be removed but will increase the rlimit

Marsman1996: I set the 0 increase when running the strip skill

rikosellic: We can also simplify this after I think the definitions are good.

rikosellic: Besides, I think increasing `rlimit` is reasonable here, because this introduces new proofs that include triggers. What we should avoid is a single function that requires `verifier::rlimit` change and runs for a very long time. It should be okay to add new proofs without `rlimit` changes, since Verus verification runs in parallel. Functions that run in a reasonable time will hardly increase the overall verification time. 

In principle, if proving a function does not require `rlimit`, then shorter is better.

verified_libs/vstd_extra/src/bits.rs
Marsman1996: Without this, there will be multiple errors:
```
error: assertion failed
   --> ostd/src/cpu/set.rs:224:27
    |
224 |     assert forall|j: int| !set@.contains(j) by {
    |                           ^^^^^^^^^^^^^^^^^ assertion failed

error: postcondition not satisfied
   --> ostd/src/cpu/set.rs:319:13
    |
319 |             ret@ == Set::empty(),
    |             ^^^^^^^^^^^^^^^^^^^^ failed this postcondition
...
324 |         Self::with_capacity_val(num_cpus(), 0)
    |         -------------------------------------- at the end of the function body

error: postcondition not satisfied
   --> ostd/src/cpu/set.rs:320:13
    |
320 |             ret.inv(),
    |             ^^^^^^^^^ failed this postcondition
...
324 |         Self::with_capacity_val(num_cpus(), 0)
    |         -------------------------------------- at the end of the function body

error: postcondition not satisfied
   --> ostd/src/cpu/set.rs:603:13
    |
603 |               final(self)@ == Set::empty(),
    |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ failed this postcondition
...
608 | /         proof! {
609 | |             let seq = smallvec_view(&self.bits);
610 | |             assert(forall|k: int| 0 <= k < seq.len() ==> seq[k] == 0u64);
611 | |         }
    | |_________- at the end of the function body

error: postcondition not satisfied
   --> ostd/src/cpu/set.rs:604:13
    |
604 |               final(self).inv(),
    |               ^^^^^^^^^^^^^^^^^ failed this postcondition
...
608 | /         proof! {
609 | |             let seq = smallvec_view(&self.bits);
610 | |             assert(forall|k: int| 0 <= k < seq.len() ==> seq[k] == 0u64);
611 | |         }
    | |_________- at the end of the function body
```

ostd/src/cpu/set.rs
Marsman1996: use `type_invariant ` here need to add more `no_unwind` and here
```
    422 -        if part_idx >= self.bits.len() {
    423 -            self.bits.resize(part_idx + 1, 0);
```
`SmallVec::resize` might panic, but `type_invariant` needs it to be `no_unwind`

And in `new_full` and `add_all`, they will temporarily break the inv.

Marsman1996: maybe weak the inv here

verified_libs/vstd_extra/src/bits.rs
rikosellic: We should add a more general lemma and manually call it at these places.

ostd/src/cpu/set.rs
rikosellic, line 1: In spec mode, call `bit_idx` and `part_idx` will work, no need to use `bit_idx_spec`, part_idx_spec``

Marsman1996: `part_idx` and `bit_idx` take `CpuId`, while `bit_at` uses an `int` that may represent padding bits outside
  the valid CPU range. Since `when_used_as_spec` requires matching signatures, reusing them here would
  require changing executable code or the spec model. Therefore, the separate spec helpers are
  necessary.

verified_libs/vstd_extra/src/bits.rs
Marsman1996: without broadcast, we have to add `let ret = ` in `new_empty`

ostd/src/cpu/set.rs
rikosellic: Indeed, makes sense.

ostd/src/cpu/set.rs
```diff
@@ -9,143 +25,854 @@ use crate::const_assert;
...
+#[verifier::allow(autoderive_clone_without_spec)]
+#[verus_verify]
 pub struct CpuSet {
     // A bitset representing the CPUs in the system.
     bits: SmallVec<[InnerPart; NR_PARTS_NO_ALLOC]>,
 }
 
 type InnerPart = u64;
 
-const BITS_PER_PART: usize = core::mem::size_of::<InnerPart>() * 8;
```
rikosellic, line 19: We should preseve this as a comment.

ostd/src/cpu/set.rs
```diff
@@ -9,143 +25,854 @@ use crate::const_assert;
...
+}
+
+impl CpuSet {
+    /// Number of set bits in the backing words.
+    pub closed spec fn count_spec(&self) -> int {
+        count_set_bits(smallvec_view(&self.bits))
+    }
+}
+
+proof fn lemma_cpucount_fits()
```
rikosellic: Auxiliary proof functions like this (in particular, private ones) can be moved to the end of the file, so that we can focus on APIs and critical proofs.

ostd/src/cpu/set.rs
```diff
@@ -9,143 +25,854 @@ use crate::const_assert;
...
+                j,
+            )
+    }
+}
+
+/// A CPU set whose backing words are all zero has an empty abstract view.
+proof fn lemma_empty_bits_imply_empty_set(set: &CpuSet)
+    requires
+        forall|i: int|
+            0 <= i < smallvec_view(&set.bits).len() ==> smallvec_view(&set.bits)[i] == 0u64,
```
rikosellic: Will it be better to use `Seq::all`?

rikosellic (changes requested):

ostd/src/cpu/set.rs
Marsman1996: We have a comment here:
```rust
// Original exec: `const BITS_PER_PART: usize = core::mem::size_of::<InnerPart>() * 8;`
// literalized to `64` inside `verus!` so Verus sees a known value (exec `% BITS_PER_PART` ↔ spec `% 64`).
```

rikosellic: Indeed, my fault,

ostd/src/cpu/set.rs
```diff
@@ -9,143 +25,690 @@ use crate::const_assert;
...
+    vstd::set::group_set_lemmas,
+    vstd::set_lib::range_set_properties,
+};
+
+/// `div_ceil(n, BITS_PER_PART)`: the number of 64-bit words needed to hold `n` bits.
+pub closed spec fn parts_for_cpus_spec(n: usize) -> usize {
+    if n == 0 {
+        0
+    } else {
+        (((n as int) + 63) / 64) as usize
```
rikosellic: The as int may be redundant.

ostd/src/cpu/set.rs
```diff
@@ -9,143 +25,690 @@ use crate::const_assert;
...
+{
+    if end <= 0 {
+        0
+    } else {
+        count_set_bits_prefix(seq, end - 1) + u64_set_bits(seq[end - 1])
+    }
+}
+
+/// Number of set bits in all words of `seq`.
+pub open spec fn count_set_bits(seq: Seq<u64>) -> int {
```
rikosellic: To be more idiomatic, one can use subrange and fold_left

rikosellic (changes requested):

ostd/src/cpu/set.rs
```diff
@@ -157,7 +715,9 @@ pub struct AtomicCpuSet {
 }
 
 type AtomicInnerPart = AtomicU64;
-const_assert!(core::mem::size_of::<AtomicInnerPart>() * 8 == BITS_PER_PART);
+// Original exec: `const_assert!(core::mem::size_of::<AtomicInnerPart>() * 8 == BITS_PER_PART);`
+// (`const_assert!` expands to exactly this `const _: () = assert!(..)` item.)
+const _: () = assert!(core::mem::size_of::<AtomicInnerPart>() * 8 == BITS_PER_PART);
```
rikosellic: This is pretty weird, should we just comment this?

ostd/src/cpu/set.rs
```diff
@@ -9,143 +25,685 @@ use crate::const_assert;
...
+    (i % 64) as usize
+}
+
+/// Number of set bits in all words of `seq`.
+pub open spec fn count_set_bits(seq: Seq<u64>) -> int {
+    seq.fold_left(0, |count: int, word: u64| count + u64_set_bits(word))
+}
+
+/// Expected value of word `idx` in a full CPU set.
+pub open spec fn full_set_word(num_cpus: int, len: int, idx: int) -> u64 {
```
rikosellic: What does this function mean? Should it be a `pub open spec fn`?

rikosellic (changes requested):

ostd/src/cpu/set.rs
Marsman1996: agree

ostd/src/cpu/set.rs
Marsman1996: `full_set_word` models the expected backing word for a full `CpuSet`: every word is `u64::MAX`, except a partially used final word, where only the low `num_cpus % 64` bits are set. Its callers establish that `len` is the number of words and `idx` is in range.

And it shall not be `pub open`.

ostd/src/cpu/set.rs
```diff
@@ -9,143 +25,685 @@ use crate::const_assert;
...
     cpu_id.as_usize() / BITS_PER_PART
 }
 
+#[verus_verify]
+#[verus_spec(
+    returns bit_idx_spec(cpu_id@),
+)]
 const fn bit_idx(cpu_id: CpuId) -> usize {
+    proof! {
+        reveal(bit_idx_spec);
```
rikosellic: Is this reveal necessary? I believe `bit_idx_spec` should be visible in this module where it is defined.

Marsman1996: Will run simplify eventually

ostd/src/cpu/set.rs
```diff
@@ -9,143 +25,685 @@ use crate::const_assert;
...
+    (i % 64) as usize
+}
+
+/// Number of set bits in all words of `seq`.
+pub open spec fn count_set_bits(seq: Seq<u64>) -> int {
+    seq.fold_left(0, |count: int, word: u64| count + u64_set_bits(word))
+}
+
+/// Expected value of word `idx` in a full CPU set.
+spec fn full_set_word(num_cpus: int, len: int, idx: int) -> u64 {
```
rikosellic: Since this is an auxiliary function only used in proofs, we can also throw this to the end of the file.

ostd/src/cpu/set.rs
```diff
@@ -9,143 +25,685 @@ use crate::const_assert;
...
+    if idx == len - 1 && bit_idx_spec(num_cpus) != 0 {
+        ((1u64 << bit_idx_spec(num_cpus)) - 1) as u64
+    } else {
+        !0u64
+    }
+}
+
+/// Bit `i` is set in the bit sequence `seq`.
+pub open spec fn bit_at(seq: Seq<u64>, i: int) -> bool {
```
rikosellic: This function can also be a `closed spec fn`, it only reflects the internal implementation of `CpuSet` and is not used in `CpuSet`'s API spec.

ostd/src/cpu/set.rs
```diff
@@ -9,143 +25,685 @@ use crate::const_assert;
...
+            assert(p < smallvec_view(&self.bits).len());
+            assert(0 <= b < 64);
+        }
         if part_idx >= self.bits.len() {
             self.bits.resize(part_idx + 1, 0);
         }
-        self.bits[part_idx] |= 1 << bit_idx;
+        // Original exec: `self.bits[part_idx] |= 1 << bit_idx;`
+        // routed via `as_mut_slice()` — SmallVec's own `IndexMut` impl has no Verus model.
+        self.bits.as_mut_slice()[part_idx] |= 1 << bit_idx;
```
rikosellic: Can we add a model for `SmallVec`'s `IndexMut`?

ostd/src/cpu/set.rs
```diff
@@ -9,143 +25,685 @@ use crate::const_assert;
...
+    )]
     pub fn contains(&self, cpu_id: CpuId) -> bool {
+        proof! {
+            use_type_invariant(&cpu_id);
+        }
         let part_idx = part_idx(cpu_id);
         let bit_idx = bit_idx(cpu_id);
-        part_idx < self.bits.len() && (self.bits[part_idx] & (1 << bit_idx)) != 0
+        // Original exec: `self.bits[part_idx]`
+        // routed via `as_slice()` — SmallVec's own `Index` impl has no Verus model.
```
rikosellic: Can we add a model for `Index`?

rikosellic (changes requested):

ostd/src/cpu/set.rs
Marsman1996: all removed by the strip skill

ostd/src/cpu/set.rs
rikosellic, line 1: This pattern `u64 & (1 << b) != 0`, i,e,, checking whether the bit is set at position `b` occurs many times in this PR. Can you define it as a spec function? This can be reused many times, including in the specification of `count_ones`.

rikosellic: Well, I have not figured out the best way yet. One option is to define it with `nat` and `pow2` like `low_bit_mask`, supporting an arbitrary range. Another option is to use bit operations and define it separately for `u8` ... `usize` with a macro.

Marsman1996: Done with the bit-ops macro

verified_libs/vstd_extra/src/bits.rs
```diff
@@ -0,0 +1,172 @@
...
+define_bit_is_set!(u16_bit_is_set, u16, 1u16);
+define_bit_is_set!(u32_bit_is_set, u32, 1u32);
+define_bit_is_set!(u64_bit_is_set, u64, 1u64);
+define_bit_is_set!(u128_bit_is_set, u128, 1u128);
+define_bit_is_set!(usize_bit_is_set, usize, 1usize);
+
+verus! {
+
+/// Every bit of the all-ones word is set.
+pub broadcast proof fn lemma_u64_allones_bit(k: int)
```
rikosellic: The lemmas can also be generalized to other types. And `bit_is_set` is not used in these lemmas' conditions.

verified_libs/vstd_extra/src/external/bits.rs
```diff
@@ -0,0 +1,90 @@
...
+verus! {
+
+/// The number of set bits among the lowest `n` bits of `w`.
+spec fn u64_set_bits_rec(w: u64, n: u64) -> int
+    decreases n,
+{
+    if n == 0 {
+        0
+    } else {
+        u64_bit_is_set(w, 0) as int + u64_set_bits_rec(w >> 1u64, (n - 1) as u64)
```
rikosellic: This definition is correct, but not very straightforward. Is it clearer to define it as 
```
 u64_bit_is_set(w, n) as int + u64_set_bits_rec(w , (n - 1) as u64)
```

rikosellic (approved):
LGTM now, thanks for your contributions!

PR #774, opened 2026-09-16
Fix irc11 verus upgrade

hiroki-chen:


hiroki-chen:
Might need an autoupdater or something

PR #775, opened 2026-09-17
prove tracked_new_val in vstd_extra

DID-Lab-SZU:


verified_libs/vstd_extra/src/ghost_tree.rs
```diff
@@ -593,16 +593,36 @@ impl<T: TreeNodeValue<L>, const N: usize, const L: usize> TreeNode<T, N, L> {
...
             res.inv(),
         returns
             Self::new_val(val, lv),
-    ;
+    {
+        let tracked children = Self::tracked_none_seq(N as nat);
+        Self::tracked_new(val, lv, children)
+    }
+
+    proof fn tracked_none_seq(n: nat) -> (tracked res: Seq<Option<Self>>)
```
rikosellic: I think we can inline this function in `tracked_new_val` since it is used only once.

rikosellic:
This PR fails on MacOS. I don't know why. It is even not in a same crate.

rikosellic:

PR #778, opened 2026-09-17
prove: boot::memory_region::as_aligned

Marsman1996:
```
/// Returns a page-aligned version of the region.
///
/// `Usable` regions are aligned inward (the result is a sub-region) and
/// the other regions are aligned outward (the result is a super-region).
///
/// # Panics
///
/// The method computes `end - base` after alignment. A `Usable` region
/// whose inward-aligned base would pass its inward-aligned end (e.g.,
/// `MemoryRegion::new(1, 1, MemoryRegionType::Usable)`) makes the
/// subtraction overflow. The upstream code has no guard for this case;
/// callers must only pass `Usable` regions that remain valid regions
/// after inward alignment, as reflected in the precondition below.
```


ostd/specs/arch/x86/mod.rs
```diff
@@ -91,6 +91,14 @@ pub proof fn lemma_max_paddr_range()
     assert(MAX_PADDR + LINEAR_MAPPING_BASE_VADDR < usize::MAX) by (compute_only);
 }
 
+pub proof fn lemma_page_size_pow2()
```
rikosellic: Well, I don't think we need this as a separate lemma. For now, just call `lemma_pow2_is_pow2_to64()` at the call site. When we move to general arch later, this property will be included in a more general arch constant requirement.

Marsman1996: By using GPT, we do not even need this lemma 👀

rikosellic: Sounds reasonable to me.

rikosellic:
Please merge the latest change.

Marsman1996:
> Please merge the latest change.

merged

ostd/specs/mm/frame/memory_region_specs.rs
```diff
@@ -22,13 +22,61 @@ impl MemRegionModel {
...
         self.end <= region.base || region.end <= self.base
     }
 
+    /// Whether the region is empty.
+    pub open spec fn is_empty(self) -> bool {
+        self.base == self.end
+    }
+
+    /// Whether `addr` is in the half-open range `[base, end)` of the region.
+    pub open spec fn contains(self, addr: int) -> bool {
```
rikosellic: Is this function used?

ostd/specs/mm/frame/memory_region_specs.rs
```diff
@@ -22,13 +22,61 @@ impl MemRegionModel {
...
+        self.base <= addr < self.end
+    }
+
+    /// Whether the region overlaps the other region.
+    pub open spec fn overlaps(self, other: Self) -> bool {
```
rikosellic: Is this function used?

ostd/specs/mm/frame/memory_region_specs.rs
```diff
@@ -22,13 +22,61 @@ impl MemRegionModel {
         self.typ == old_region.typ && old_region.base <= self.base <= self.end <= old_region.end
     }
 
+    /// The mirror of [`MemRegionModel::is_sub_region`]: the region contains
+    /// `old_region`, with the same type.
+    pub open spec fn is_super_region(self, old_region: Self) -> bool {
```
rikosellic: The `old_region` name is weird, maybe `small_region`, `sub_region` or just `region`.

rikosellic (changes requested):

ostd/specs/mm/frame/memory_region_specs.rs
Marsman1996: weird, why it keeps adding these useless functions

rikosellic:
The spec and proof lgtm, but I'm thinking do we really need a `MemRegionModel`? A `MemRegion` is characterized by its `start`, `end` and `type`, which are already provided by APIs. This is not like `CpuSet`, where we have an abstraction that is very different from the implementation.

Marsman1996:
> I'm thinking do we really need a `MemRegionModel`?

keeping it or leaving it both are ok for me.

Marsman1996:
> The spec and proof lgtm, but I'm thinking do we really need a `MemRegionModel`?

Remove the `MemRegionModel` in the latest commit

ostd/specs/mm/frame/memory_region_specs.rs
```diff
@@ -1,52 +1,20 @@
...
-    pub open spec fn is_separate(self, region: Self) -> bool {
-        self.end <= region.base || region.end <= self.base
-    }
-
-    pub open spec fn bad() -> Self {
-        MemRegionModel { base: 0, end: 0, typ: 0 }
-    }
-}
-
 pub ghost struct MemoryRegionArrayModel<const LEN: usize> {
```
rikosellic: This also seems redundant.

Marsman1996: It's the abstract view of `MemoryRegionArray`: the `Seq` of regions currently in the set (`Seq::new(count, |i| regions[i])`), hiding the fixed 512-slot storage, `count`, and the `bad()` padding. The `deref`/`push` contracts (`ret@ == self@.regions`, `old(self)@.push(region)`) are stated in terms of it, since `pub` fn contracts can't reference the private fields.

rikosellic: Can we directly view `MemRegionArray` as `Seq<MemRegion>`?

Marsman1996: Well, we can

rikosellic:

rikosellic (approved):

PR #782, opened 2026-09-17
prove: boot::memory_region::Deref

DID-Lab-SZU:


ostd/src/boot/memory_region.rs
```diff
@@ -236,6 +236,13 @@ impl<const LEN: usize> Inv for MemoryRegionArray<LEN> {
     }
 }
 
+impl<const LEN: usize> MemoryRegionArray<LEN> {
```
rikosellic, line 233: Since this is `type_invariant`, you may delete `impl Inv for MemoryRegionArray` now,

ostd/src/boot/memory_region.rs
```diff
@@ -230,10 +230,15 @@ pub struct MemoryRegionArray<const LEN: usize = MAX_REGIONS> {
...
 
-impl<const LEN: usize> Inv for MemoryRegionArray<LEN> {
-    closed spec fn inv(self) -> bool {
+impl<const LEN: usize> MemoryRegionArray<LEN> {
+    #[verifier::type_invariant]
+    closed spec fn type_inv(self) -> bool {
         self.count <= LEN
     }
+
+    pub closed spec fn inv(self) -> bool {
```
rikosellic: Do we need this? Where is it used.

ostd/src/boot/memory_region.rs
```diff
@@ -244,37 +245,38 @@ impl<const LEN: usize> View for MemoryRegionArray<LEN> {
...
 }
-/*
+#[verus_verify]
 impl<const LEN: usize> Deref for MemoryRegionArray<LEN> {
     type Target = [MemoryRegion];
 
+    #[verus_spec(ret =>
+        ensures
+            ret@.len() == self@.regions.len(),
+            forall|i: int| 0 <= i < ret@.len() ==> (#[trigger] ret@[i])@ == self@.regions[i],
```
rikosellic: Can we write a simpler postcondition here? Like `ret@ == self@.regions`

rikosellic:

PR #787, opened 2026-09-19
prove(ostd-pod): add from_bytes back and sync with mainline

Marsman1996:
Add `from_bytes` back in ostd-pod, and some other modification

refer to https://github.com/asterinas/vostd/blob/phaseII/mainline/ostd/libs/ostd-pod/src/lib.rs


rikosellic:

PR #792, opened 2026-09-20
chore: reorder import

rikosellic:
Aggregate separate `vstd::`, `vstd_extra::` and `crate::specs` into blocks and list them at the beginning of files.


Marsman1996:
Why add a blank line between `vstd::` and `vstd_extra::` 🤔

rikosellic:
> Why add a blank line between `vstd::` and `vstd_extra::` 🤔

Good point, I will delete this.

Marsman1996:
And how about the specs?

If specs are separated, maybe better to point it out in the code guideline I think.

rikosellic:
> And how about the specs?
> 
> If specs are separated, maybe better to point it out in the code guideline I think.

Indeed, the ideal style is 

```
use vstd::{...};
use vstd_extra::{...};

use crate::specs::{...};

/// Original imports in source code
use ...;
```
A blank line is added between `vstd_extra` and `crate::specs` to prevent formatter reordering. Better to record this in the coding guideline.

PR #796, opened 2026-09-21
prove: AtomicCpuSet

Marsman1996:
Have to introduce one more axiom, the reason is recorded in the comment


rikosellic:
Before reviewing this, I'm thinking about whether it would be easier to change `CpuSet`'s model from `Set<int>` to `Set<CpuId>`?

Marsman1996:
> I'm thinking about whether it would be easier to change `CpuSet`'s model from `Set<int>` to `Set<CpuId>`?

I do not think it is easier to use `Set<CpuId>`, see the latest commit 🤔

rikosellic:
Thanks! Let's just keep it as `Set<int>`.

rikosellic:
I believe there is an `AtomicCpuSet` placeholder in `specs/mm/cpu`, please also delete this and fix imports in this PR.

Marsman1996:
The whole `ostd/specs/mm/cpu.rs` is deleted

Marsman1996:
The platform-issue see https://github.com/asterinas/vostd/actions/runs/35582921672/job/106279632225, caused by https://github.com/asterinas/vostd/compare/3c08eee01e74ce08e02aafae49a96e33a8f7a88c..a5315bcb426148927d58acbbcc29caa51bfe5be8

